PR #20, opened 2017-11-16
a few examples as python scripts

sauln:
I found these examples on an old fork I had.  I honestly can't remember if I wrote them or not.  Do you know where they came from? 

Feel free to add them to the repo if you want. I've tested them and they work fine.


lmcinnes:
Greatly appreciated!

PR #30, opened 2017-11-30
Integration tests should run w/o conda also

sauln:
It looks like the travis-ci tests only run on conda distributions.  UMAP should be able to run without conda just fine.

This is a work in progress. I'm having trouble getting travis to work correctly and am not sure any better way to test it except pushing amendments to the commits.


lmcinnes:
I agree, but clearly your travis-fu is better than mine; I simply worked from available templates. Anything you can do here is greatly appreciated.

Issue #26, opened 2017-11-24
Precomputed distances


LGro:
Would you be interested in / see any obstacles for implementing UMAP on a distance matrix? After a first glance this seems to be quite straight forward to include. 
I'd be inclined to contribute.

lmcinnes:
No that's certainly quite feasible, the only catch being that you have to have computed the distance matrix first. I would be happy to work though and merge a pull request if you want to put in the work. Thanks!

LGro:
Sure, should be helpful in the cases where people have precomputed distance matrices. 
I have a working implementation already (looks very nice!) and will submit a pull request in the upcoming days.

So far I have added a `precomputed` flag to `make_nn_descent` that is set when the given metric is set to the string precomputed. However, I'm still struggling with what to do with the distance function in that case, because it seems like it expects a numba function in any case. 
Would you prefer a flag that changes the behaviour or a dummy distance function named `precomputed` that is then checked for its name in `nn_descent`? 
Maybe we can come up with an even more elegant way.

If I couldn't make clear what I mean just wait for my pull request and we can discuss the actual code.

lmcinnes:
I think adding a dummy distance function called ``precomputed`` to distance.py is probably the better way to go under those options (would likely have less code duplication).

Since you have the full distance matrix, however, could you not compute the nearest neighbors by an argsort across an index of the distance matrix? That might save you having to mess with nn_descent at all. Then again, if you have something that works I'm not really going to argue.

LGro:
You are right, that's more elegant and faster.

PR #31, opened 2017-12-01
add metric option precomputed (#26)

LGro:
I'm not sure why the metric validation in umap_.py:1308-1314 was commented out by you in da04e17875225a14edcf80e0832199a23c06fd83, but in case you want to integrate this again, you would need to add the following:
```
elif metric == 'precomputed':
    pass
```


LGro:
Sorry for falsely referencing #24 instead of #26 in the commit.

lmcinnes:
Looks great! Thanks for taking the time to do this, it is greatly appreciated. If you have any further ideas for improvements please don't hesitate to propose them. I'll gladly accept more pull requests like this.

PR #47, opened 2018-02-21
Excluded scipy v1.0.0 from setup.py

wflynny:
See issue #39.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ configuration = {
     'license' : 'BSD',
     'packages' : ['umap'],
     'install_requires' : ['scikit-learn >= 0.16',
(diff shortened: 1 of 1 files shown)


gforsyth:
I don't think this is a good way to fix this -- it will prevent people from installing `umap` if they have the latest version of SciPy.  I think @lmcinnes has a fix in master that fixes the error in #39 without pinning a version

wflynny:
@gforsyth Saying this again here (as I did in #39), this PR can stay open until after newer versions of `scipy` are released.  Or it can be closed.  Just giving @lmcinnes the option.

sleighsoft:
@lmcinnes Did this get resolved already?

Issue #46, opened 2018-02-21
Make `umap.__version__` return current version


MLWave:
For replication purposes it is useful to print/log the current version of UMAP.

Is there are variable that holds this information?

lmcinnes:
It should now ... Hopefully I can keep this all synced.

MLWave:
Thank you very much!

MLWave:
Sorry, I think:

```
__version__ = 0.2.1
```
should be:
```
__version__ = "0.2.1"
```

lmcinnes:
Good catch. Thanks.

PR #48, opened 2018-02-21
dynamically add __version__ variable based on what's in setup.py

sauln:
This is an alternative fix for #46. This will print the version number that is set in the setup file so you only have to change the version number in one place.

> Hopefully I can keep this all synced.

Now you won't have to worry about it.

### DIFF
--- a/umap/__init__.py
+++ b/umap/__init__.py
@@ -1,2 +1,4 @@
 from .umap_ import UMAP
-__version__ = "0.2.1"
+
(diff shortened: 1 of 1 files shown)


lmcinnes:
Thanks, I figured there was some way to do this, but I didn't see anything obvious that didn't involve a little bit of hackery. Presumably ``pkg_resources`` conveniently hides all that hackery for me.

PR #52, opened 2018-02-28
Added access to high-dimensional simplicial set through class UMAP

pcamara:
I also updated the documentation of `UMAP` and `simplicial_set_embedding` to account for the new parameter `negative_sample_rate`, instead of the old `n_edge_samples`.


lmcinnes:
Thanks! The doc updates are very helpful as well. Greatly appreciated.

pcamara:
It is my pleasure. Thank you for this great piece of work.

PR #60, opened 2018-03-22
add transform() method for compatibility with scikit-learn pipelines

morenobonaventura:
added transform() method for compatibility with scikit-learn pipelines

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -1478,3 +1478,14 @@ class UMAP(BaseEstimator):
         """
         self.fit(X)
         return self.embedding_
+      
+    def transform(self, X):
+      """Fit X into an embedded space and return that transformed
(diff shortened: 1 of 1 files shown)


lmcinnes:
Hi, thanks for the contribution! I actually have a transform method that can work more like PCA and other similar approaches underway in the 0.3dev branch. Given that I was planning to just wait to complete v0.3 work and then merge it all into master. I would appreciate it if you could check the 0.3dev branch and see if that would meet your needs or if you have different requirements that we can discuss.

Thanks again.

sleighsoft:
I will close this issue in favor of the code already in the master branch. Let me know if there is any problem with this. 

Nonetheless, thank you for contributing @morenobonaventura.

PR #70, opened 2018-05-20
Minor fix with defaults for weighted minkowski

yinshanyang:
Hi, I’m not completely sure if this is correct, but it seems like the default value for `w` in `weighted_minkowski` should be `_mock_ones` instead of `_mock_identity`.


lmcinnes:
You are quite correct. Good catch, thanks.

PR #103, opened 2018-08-04
Implement inverse transform

josephcourtney:
I implemented an inverse transform allowing a user to find vectors in data space corresponding to points in embedding space. The basic algorithm is the same as forward UMAP except that the simplicial set is based on a Delaunay complex rather than a Vietoris-Rips complex to provide for smooth behavior in the gaps between clusters. However, this means that inverse_transform(transform(points)) != points in general. 

It was necessary to modify optimize_layout to allow for non-Eucldiean distance metrics in the "embedding" space so that it could be used backwards. It is also now possible to choose whether the used membershhip function is for the input set or the output set. As a side effect, it is now possible to use UMAP to embed data onto a torus or a sphere (or any manifold for which you define a distance and gradient). 

Example scripts for inverse transformation and embedding onto a torus or a sphere are in the examples directory.

This implementation is still a little rough but is a proof of concept. Currently, only the euclidean and haversine metrics are supported. Functions returning the gradient of other metrics are still needed.

Embedding of the sklearn digits dataset onto a torus
torus

Embedding of the sklearn digits dataset onto a sphere
sphere

Inverse transform of a grid of points (shown as black xs) back into data space. Dataset is MNIST. Some points look like simple pixel-wise averages of neighboring points. The sharpness of generated vectors increases with higher values of n_epoch.
inverse_transform


lmcinnes:
This looks great. I'll have to take some time and try to get through reviewing it all. Unfortunately that may not happen until after I get back from vacation.

josephcourtney:
Thanks. I see that my changes have caused some tests to fail and I have some PEP8 violations. I haven’t used GitHub all that much; should I make changes to fix those problems and make a new pull request or can I amend the current one?

lmcinnes:
Just continue to push to the branch of yours that you are trying to merge and they will appear here -- if that fixes the issues then great. Don't worry too much about the pep8 violations. I need to get around to tweakijng the setting on it so it isn't quite as strict with whitespace/line length issues.

lmcinnes:
So, first of all, this looks fantastic. I really like the way you've structured this with a nice separation of concerns. I'll try to go through and review the changes with specific comments at various points.

The CI failure is largely due to one line that I believe I can see how to fix (it will require some deeper changes, but they were going to have to happen anyway). I would be interested in some before/after benchmarking of runtime performance: I'm keen to keep roughly the same performance for the basic use case if at all possible.

josephcourtney:
It is definitely difficult straddling the gap between performance and good modularization. This was my first time working with Numba so I just followed your examples of using a "make_function" as a factory for a numba-optimized, specialized version of the function.

I wasn't sure of how to implement the metric+gradient functions. There is usually a big benefit to calculating the distance and its gradient at the same time since you can reuse many intermediate values but since numba requires (?) that the output type of the function doesn't change based on arguments, I had to write two versions, one that returns the gradient and one that doesn't. I am not happy with the logic I wrote to retrieve the distance+gradient function but I couldn't come up with a good solution at the time.

One clean way to get around that issue is to use an automatic differentiation library like autograd, tensorflow, pytorch, etc. but that would essentially be a complete rewrite of UMAP which is definite overkill.

One other concern is how I managed the sigma and rho values. I recalculated them and passed them  to optimize_layout but I'm pretty sure that is not the best way to do it, especially since, in the reverse direction, the set of neighbors is derived from a Delaunay complex rather than a Vietoris-Rips complex. I'm not sure how to approach that as my math education beyond basic ODEs and PDEs is thin and self taught. I'm just now making my way through a topology textbook.

I will do some basic benchmarking and post results here.

Thank you for creating UMAP and being so supportive of its further development. Digging into this project has been very enjoyable.

umap/umap_.py
```diff
@@ -677,152 +703,207 @@ def rdist(x, y):
...
-                        a * pow(dist_squared, b) + 1
+                    dist_output, grad_dist_output = output_metric(
+                        current, other, *output_metric_kwds
                     )
 
-                    if not np.isfinite(grad_coeff):
-                        grad_coeff = 4.0
+                    w_l = pow((1 + a * pow(dist_output, 2 * b)), -1)
+                    if inverse:
+                        grad_coeff = (
```
lmcinnes: I admit this is not obvious to me -- can you sketch a derivation for this as the appropriate gradient? I would have thought that ``rhos[j]`` ought to also play a role?

umap/umap_.py
```diff
@@ -1631,6 +1777,149 @@ def transform(self, X):
...
+                    "output_metric is neither callable, " + "nor a recognised string"
+                )
+
+
+        optimize_layout = make_optimize_layout(
+            _input_distance_func,
+            tuple(self._input_metric_kwds.values()),
+        )
+
+        sigmas, rhos = smooth_knn_dist(
```
lmcinnes: I suspect this is not what you want here, unless the sigmas are playing a different role than I would expect?

lmcinnes:
I actually like the approach you took with separate gradient aware distance functions -- it is nicely numba compatible, and only needed for cases where we want to use a particular distance as an output metric. The whole thing seemed pretty clean to me, and easy to extend which is also important (I'm looking forward to adding a Poincare disc hyperbolic distance as an optional output distance).

I think the correct way to handle the sigmas and rhos is outlined in some of the comments above, but explicitly I think the correct approach is return them with the fuzzy simplicial set and store them as (private) attributes for later (re)use.

lmcinnes:
Hi @josephcourtney, are you still working on this? I'm very keen to get it added to the library as it provides some very powerful features and looks like a generally excellent implementation. If you don't have time I will be happy to merge this into 0.4dev and work through the issues myself. Let me know either way, and thanks again for all your work on this, it's amazing.

lmcinnes:
That's great news. I was not trying to rush you, but rather just wanted to check if you had moved on, in which case I definitely wanted to pick up from what you've already done. As long as you are still chipping away I am more than happy to patiently wait. Thanks again for the contribution!

josephcourtney:
My perfectionist tendencies have gotten the better of me. I was unsatisfied that inv(inv(transform)) != transform due to my choice to use the Delaunay complex. Happily, this thesis (https://esc.fnwi.uva.nl/thesis/centraal/files/f1267161058.pdf), as noted in @lmcinness tweet (https://twitter.com/leland_mcinnes/status/1044235347125841920) has come along and has ensured that inv(inv(transform)) == transform by building an autoencoder. 

Since a better approach has been presented, I will fix up problems in my implementations of the supporting functions (enhancements to optimize_layout, gradient implementations for metric functions, etc.) and commit them. The real question is whether the Delaunay complex implementation is worth including as "the" inverse transform. Maybe it would be more appropriate as an example script of how to use the new optimize_layout features rather than as a core feature of UMAP?

lmcinnes:
I actually really like the approach you used here -- and personally prefer it to an autoencoder approach as the "correct" approach to a direct version of inverse transform for UMAP. I remain unconcerned about a difference between inv(inv(transform)) and transform, particularly as both are stochastic, at least for now. I definitely want to see this merged, and continue to plan to offer it as the inverse_transform method available in the 0.4 release of UMAP, so please, by all means finish up this implementation! I looks forward to it.

josephcourtney:
I think the major problems are fixed. I've attached an explanation of the new gradient update equations in optimize_layout ([derivation.pdf](https://github.com/lmcinnes/umap/files/2431854/derivation.pdf)). I think I've fixed all the problems I introduced and brought it up to compatibility with the 0.4dev branch. Once we agree that the code is correct I can work on some optimizations.

The Codacy issues seem trivial to me. The syntax for numpy.mgrid is out of our control and the clarity of the code with unused variables is worth it to me.

I also have written gradients for a large portion of the metrics but I haven't written any tests (finite differences) to confirm they are correct.

Current TODOs as I see them:
- write tests for metric gradients
- optimize Delaunay complex creation
- change some hard-coded choices (minimum number of neighbors in Delaunay complex) into parameters
- write a notebook / blog post explaining inverse transform and demonstrating it on multiple datasets
- write sparse versions of metric gradients
- implement optimization method for non-differentiable metrics

lmcinnes:
Don't stress about the codacy stuff. I had it look at umap generally just to see if it was going to be useful. In general I haven't found it as useful as I had hoped (and some false positive issues). I was planning to turn it off.

josephcourtney:
I already found a bug in the inverse transform implementation. The weights of the 1-skeleton of the Delaunay complex were not L1 normalized before calculating the init_transform. The samples were visually reasonable but they were FAR outside the data manifold (pixel values of 1e4) so any optimization was uniformly scaling the pixel values to pull them back towards the manifold, not pulling them sideways towards other input data points. 

After fixing that the inverse transformed points look similar but have pixel values in the range 0-255. Also, if I add substantial pixel noise before calling optimize_layout the resulting output does show the expected behavior that points that start in dense clusters have a large reduction in noise while points that start between clusters (where attractive forces are lower) stay noisy.

I'm going to perform a few more tests but it seems to fix the problem. I'll push a commit shortly.

lmcinnes:
I looked through everything and it all looks good now. The TODO list you provide above looks good, although the last couple of items I would consider as 'nice to have', and can always be dealt with later -- an error that inverse transform doesn't support sparse input or certain metrics is more than adequate at this point. Thanks for all the hard work on this. Let me know if and when you'd like to see it merged, because I am essentially ready when you are!

josephcourtney:
I think now is a good time to merge it. I can work on tests and optimizations later but I don't want the branches to diverge too much if I don't get back to working on UMAP for a week or two. The other ideas on the TODO list would be more appropriate as separate pull requests anyways.

lmcinnes:
This was a major chunk of work that provides some very powerful new features, so thanks so much for this. I'm going to have more time in November and hopefully I can get to work on 0.4 and see if I can get the release out the door so we can get your work in front of more people.

josephcourtney:
It was a lot of fun. I had no prior experience with topology but UMAP provided the motivation necessary to really dig in. Working to really understand UMAP and reading some other topological data analysis papers has given me a ton of ideas so hopefully you will be seeing some experiments or more pull requests from me in the future.

lmcinnes:
I look forward to it!

Issue #128, opened 2018-08-22
Documentation on sklearn pipelines


lmcinnes:
As per @terrytangyuan in openjournals/joss-reviews#861 , it would be beneficial if there was at least a brief section in the documentation outlining the use of UMAP with sklearn pipelines, and demonstrating using it with, for example, a clustering method, or a classifier, and potentially using GridSearchCV for hyper-parameter optimization.

johannfaouzi:
Hi,

Does the following piece of code fit your expectations? It shows that UMAP can be used as a feature extraction method to improve the accuracy on a classification task and uses several tools from scikit-learn.

```python
from sklearn.datasets import make_classification
from sklearn.model_selection import train_test_split, GridSearchCV
from sklearn.pipeline import Pipeline
from sklearn.svm import LinearSVC
from umap import UMAP


# Make a toy dataset
X, y = make_classification(n_samples=1000, n_features=300, n_informative=250,
                           n_redundant=0, n_repeated=0, n_classes=2,
                           random_state=1212)

# Split the dataset into a training set and test set
X_train, X_test, y_train, y_test = train_test_split(
    X, y, test_size=0.2, random_state=42)

# Classification with a linear SVM
svc = LinearSVC(dual=False, random_state=123)
params_grid = {"C": [10**k for k in range(-3, 4)]}
clf = GridSearchCV(svc, params_grid)
clf.fit(X_train, y_train)
print("Mean accuracy on the test set with raw data: {:.3f}".format(
       clf.score(X_test, y_test)))

# Transformation with UMAP followed by classification with a linear SVM
umap = UMAP(random_state=456)
pipeline = Pipeline([("umap", umap),
                     ("svc", svc)])
params_grid_pipeline = {"umap__n_neighbors": [5, 20],
                        "umap__n_components": [15, 25, 50],
                        "svc__C": [10**k for k in range(-3, 4)]}

clf_pipeline = GridSearchCV(pipeline, params_grid_pipeline)
clf_pipeline.fit(X_train, y_train)
print("Mean accuracy on the test set with UMAP transformation: {:.3f}".format(
       clf_pipeline.score(X_test, y_test)))
```
which outputs:
```
Mean accuracy on the test set with raw data: 0.755
Mean accuracy on the test set with UMAP transformation: 0.935
```

lmcinnes:
Looks good! If you want to add it to the examples directory, following the
examples there of the comment block at the top with a markdown explanation
I would be very happy to include it. Thanks!

PR #130, opened 2018-08-23
Add example with sklearn pipeline #128

johannfaouzi:


lmcinnes:
Excellent! Thank you as always,

PR #151, opened 2018-10-06
Change LICENSE.txt to have the actual BSD 3-Clause License

bskaggs:
The source code and README.rst all indicate the BSD 3-Clause License should be used, but the LICENSE.txt file itself is the BSD 2-Clause License.  The PR makes the LICENSE.txt file match the code comments.


lmcinnes:
Thanks, may bad.

PR #156, opened 2018-10-27
Add readme style edits

mwiens91:
Most of these are small changes, among which we have

- renaming ReadTheDocs to Read the Docs (I've never seen the former, and
  the latter is how the platform self-refers)
- using proper em-dashes ("—"s) instead of " -- "s
- adding commas in places that feel natural


sleighsoft:
It has been a while... Any reason why this did not get merged @lmcinnes ?

lmcinnes:
Simply slipped through the cracks would be my guess.

sleighsoft:
Looks good to me

PR #199, opened 2019-02-03
[Fixes #198] x and y are required to be the same length

### LINKED ISSUE #198: x and y in fit() are not required to be the same length
I just ran some code that basically boils down to:

```python
x = np.random.uniform(0, 1, (256, 10))
y = np.random.randint(10, size=(257, 1))
out = umap.UMAP().fit_transform(x, y)
```

And it took me a while to realize I made a bit mistake (the x and y weren't matched with each other) because the output still looked sort of plausible. It could be really good to add a `if y is not None and len(x) == len(y)` exception to the `fit()` and `fit_transform()` functions, just to help out other folks making similar dumb mistakes.

fiza1160:
This PR is related to #198.
I hope I got it right 😄

### DIFF
--- a/umap/tests/test_umap.py
+++ b/umap/tests/test_umap.py
@@ -732,3 +732,31 @@ def test_umap_bad_parameters():
         False,
         np.random,
     )
+
+
+def test_umap_fit_params():
+    # x and y are required to be the same length
+    u = UMAP()
+    x = np.random.uniform(0, 1, (256, 10))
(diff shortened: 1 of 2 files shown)


lmcinnes:
Looks good. Thanks for adding the tests as well -- that is certainly always appreciated.

lmcinnes:
Please feel free to submit PRs for anything else you see.

PR #215, opened 2019-04-03
Make Travis check formatting with Black

tomwhite:
Black has been used to format the source previously. This adds a check and fails the build if the source is not correctly formatted. PR authors can fix the formatting by running `black umap`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,3 +35,18 @@ are fixing a new issue feel free to file an issue and then reference it in the P
 You can [browse open issues](https://github.com/lmcinnes/umap/issues), 
 or consult the [project roadmap](https://github.com/lmcinnes/umap/issues/15), for potential code
 contributions. Fixes for issues tagged with 'help wanted' are especially appreciated.
+
+### Code formatting
+
+If possible, install the [black code formatter](https://github.com/python/black) (e.g.
+`pip install black`) and run it before submitting a pull request. This helps maintain consistency
+across the code, but also there is a check in the Travis-CI continuous integration system which
+will show up as a failure in the pull request if `black` detects that it hasn't been run.
+
+Formatting is as simple as running:
+
+```bash
+black .
+```
+
+in the root of the project.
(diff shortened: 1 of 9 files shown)


tomwhite:
This is failing because `black` isn't being installed, maybe because the build environment is cached?

lmcinnes:
I'm not really sure -- I have never been that good at this CI stuff. I'll try to look into it when I get some time; I'm kind of swamped with things right now.

jamesmyatt:
If you want to use black, then you should relax the line limit to 88: https://github.com/python/black#line-length.

jamesmyatt:
Try adding "black" to the dependencies in https://github.com/lmcinnes/umap/blob/b3816672c74f7963a79ad69186da60469484c14c/ci_scripts/install.sh#L29

lmcinnes:
Thanks @jamesmyatt. I'll see if I can get that sorted soon.

tomwhite:
Thanks for the suggestion @jamesmyatt. It's getting a bit further now, but is failing one build as `black` isn't available on Python 2, and another build because the build environment is different in some way. This will need a bit more work...

lmcinnes:
We should just start ignoring python2. It is going to be end of life soon, and I have no particular desire to support it (and have pledged to drop support for it by Jan 1 2020 anyway, so we certainly do not ahve to provide any support after that).

jlmelville:
The `black` test only needs to be carried out on one of the matrix entries, so one solution is only carrying it out on the coverage entry: it's already got its own branch of the bash script so there's no extra complexity added. 

If you go this way, I also suggest doing the black check before the tests, rather than after. This way the check fails faster. Also the coverage checks often falsely fail the build due to negligible negative changes in coverage which would otherwise hide a "true" failure due to a forgotten reformatting.

Finally, I added a section about using `black` before submitting a PR to https://github.com/lmcinnes/pynndescent/blob/master/CONTRIBUTING.md -- it would probably be helpful to add that to this PR too (everything else in the pynndescent document is copy-and-pasted from UMAP, so this is a trivial change).

tomwhite:
Thanks for the suggestions James. Here's another attempt.

lmcinnes:
Sorry for being slow to get to this, but thanks for all the work people have put it. It is definitely appreciated. Merging now, and thanks again!

PR #234, opened 2019-05-10
Fix Spivak reference in "How UMAP works"

adelejackson:
Let me know if this should go into master rather than 0.4dev.

There's also a formatting issue with the one use of latex in "How UMAP works" (line 405/406), but I'm not familiar enough with reStructuredText to know how to fix it.


lmcinnes:
Thank you!

PR #236, opened 2019-05-15
Use pynndescent dependency

tomwhite:
This is for illustration/discussion, it's not intended to be merged!

It shows the changes needed to switch to a dependency on pynndescent. Mainly it is removing the duplication between the two projects.

Still to do (not exhaustive):
* It might be worth moving the case for `metric='precomputed'` to pynndescent.
* The code for sizing `n_trees` and `n_iters` could live in pynndescent too. (It has an associated comment saying it is a hack - are there any improvements that could be made to these functions?)
* A new release of pynndescent is needed.

### DIFF
--- a/ci_scripts/install.sh
+++ b/ci_scripts/install.sh
@@ -54,4 +54,5 @@ if [[ "$DISTRIB" == "conda" ]]; then
   python setup.py develop
 else
   pip install -e .
+  pip install pynndescent # test with optional pynndescent dependency
 fi
--- a/umap/tests/test_umap.py
+++ b/umap/tests/test_umap.py
@@ -762,7 +762,7 @@ def test_umap_transform_on_iris():
     trust = trustworthiness(new_data, embedding, 10)
     assert_greater_equal(
         trust,
-        0.89,
+        0.88,
         "Insufficiently trustworthy transform for" "iris dataset: {}".format(
             trust),
     )
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -204,86 +204,107 @@ def nearest_neighbors(
 
         rp_forest = []
     else:
-        if callable(metric):
-            distance_func = metric
-        elif metric in dist.named_distances:
-            distance_func = dist.named_distances[metric]
-        else:
-            raise ValueError("Metric is neither callable, " + "nor a recognised string")
+        try:
+            # Use pynndescent, if installed (python 3 only)
+            from pynndescent import NNDescent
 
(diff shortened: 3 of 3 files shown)


tomwhite:
The Python 3 build is working, but the Python 2 one is not since pynndescent is Python 3 only, and the conda one is not since pynndescent is not in the conda repo.

lmcinnes:
Well I can fix one of those shortly. As for Python 2 ... it is not my plan to continue supporting it.

tomwhite:
Thanks @lmcinnes.

tomwhite:
@lmcinnes depending on when you plan to drop Python 2 support, one option would be for Python 2 to use the existing NN-descent code in UMAP, and for Python 3 to use the new pynndescent dependency.

lmcinnes:
I have committed to the fact that Python 2 support is getting completely dropped for this package by Jan 1 2020 *at the latest*, so I don't see any reason to jump through hoops for it -- I'm happy accepting that we can drop Python 2 when this gets merged.

Speaking of which -- you started out with this as a proposal, not to be merged, but this looks pretty good to me now. I'm retargeting it against 0.4dev, since I really do hope to get the last bits and pieces for that completed soon (ideally before SciPy this year).

tomwhite:
I think this could be merged now - it started off as an experiment, but the latest incarnation will only use the pynndescent dependency if it's been installed. That means it's also independent of the decision to drop Python 2 support.

lmcinnes:
I removed the conflicts with 0.4dev -- presuming tests pass I'll try to get this merged in.

tomwhite:
Some of the failures seem to be permissions errors. Might be worth retrying the build to see if that helps.

lmcinnes:
Sorry, I've been busy with many other things lately. It looks like the pynndescent dependency is not getting installed -- not entirely sure why.

tomwhite:
Thanks @lmcinnes. By design, the pynndescent is only being installed for the pip build, it is not installed under conda so that we can test that pynndescent is not required, and that the existing UMAP NNDescent code is used.

The failure is due to differences between master and 0.4dev, so I've opened https://github.com/lmcinnes/umap/pull/278 with a fresh commit on top of the latest 0.4dev branch.

lmcinnes:
Thanks, I've merged the other. It is greatly appreciated. Sorry for being so slow with all of this.

tomwhite:
Thanks @lmcinnes!

tomwhite:
Closing this since it has been merged into the 0.4dev branch.

PR #237, opened 2019-05-16
Fix deprecation warning with scikit-learn 0.21

jamesmyatt:
sklearn.externals.joblib is deprecated in 0.21 and will be removed in 0.23.

This change uses the joblib library directly, if available; Otherwise it falls back to sklearn.externals.joblib, which seems like the least intrusive way of making this change.

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -13,9 +13,9 @@ from sklearn.preprocessing import normalize
 from sklearn.neighbors import KDTree
 
 try:
-    from joblib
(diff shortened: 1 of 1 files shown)


jamesmyatt:
Failure in https://travis-ci.org/lmcinnes/umap/jobs/533380248 is unrelated to this PR.

lmcinnes:
Thanks -- I was just dealing with this in hdbscan, but hadn't looked at umap recently.

PR #239, opened 2019-05-19
demo umap docs build with sktda docs

sauln:
This conf file demonstrates how umap docs page would integrate the sktda docs system.

You'll see it massively simplifies the conf.py, putting the logic inside the [sktda_docs_config package](https://pypi.org/project/sktda-docs-config/) and on [github](https://github.com/scikit-tda/sktda-docs-config).

To setup and see how it runs, install `sktda-docs-config` and `sphinx_gallery` into an environment with umap. Then run `make html` in the `doc` directory. You can view the results by opening `_build/html/index.html` in your browser.

This does not include the ability to manage versions in the docs, unfortunately, but it looks like UMAP is not currently using that feature of readthedocs.

Let me know what you think and we can figure out the next steps for integration.


lmcinnes:
Thanks for this -- I'll try to look through it shortly when I have  a little more time.

lmcinnes:
Apparently time is at a premium. Perhaps we can discuss this at SciPy this year?

lmcinnes:
Sorry -- I just never quite got to this :-(

sauln:
Totally understand.  The docs are fine as they are. I've actually been thinking of converting the rest of the scikit-tda docs to readthedocs so management is a bit easier.

PR #240, opened 2019-05-20
Use random_state param (#238)

### LINKED ISSUE #238: Random state parameter is ignored
Hi,

in this line:

https://github.com/lmcinnes/umap/blob/7041502068e7bc3aa28fc00a12fffa66435410dd/umap/umap_.py#L981

shouldn't it be `random_state.normal`  instead of `np.random.normal`? The random_state parameter is ignored when init is not random/spectral string.

candalfigomoro:
Use random_state param instead of np.random, fix #238.

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -978,7 +978,7 @@ def simplicial_set_embedding(
                 tree = KDTree(init_data)
                 dist, ind = tree.query(init_data, k=2)
                 nndist = np.mean(dist[:, 1])
-                embedding = init_data + np.random.normal(
+                embedding = init_data + random_state.normal(
                     scale=0.001 * nndist, size=init_data.shape
                 ).astype(np.float32)
             else:


candalfigomoro:
I've seen it has been fixed in v0.4 https://github.com/lmcinnes/umap/commit/4cd67e52cfe254c98cacc4a20e0fbf6a96379a65

@lmcinnes 
Is v0.4 the next release or we will have a new 0.3.x release?

lmcinnes:
I suspect I'll have another 0.3.x release before 0.4 eventually makes it out the door.

candalfigomoro:
@lmcinnes 
I've seen there are others "reproducibility" fixes in v0.4 (also related to numba etc.). Could you backport all the "reproducibility" fixes into v0.3.x?

lmcinnes:
I think the other issues aren't relevant to 0.3, but I can check.

PR #248, opened 2019-06-04
Adding MacOS builds for Travis CI

vicramr:
I recently figured out how to get Travis to run tests on MacOS so I figured I'd add the functionality to umap.

I'm not 100% sure this will work; I'm mainly making this PR to trigger a Travis build so we can see what happens.

For simplicity, I've removed the check in `ci_scripts/install.sh` to use the cached version of `miniconda.sh`, because now there can be two different versions of that file. The best way to handle this would be to name them differently. That said, according to [the Travis docs](https://docs.travis-ci.com/user/caching/#things-not-to-cache), caching these downloads probably won't improve the build time. Specifically, it says "Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: python
 
+os:
+  - linux
+  - osx
+
 cache:
   apt: true
   # We use three different cache directory
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: python
 
-os:
-  - linux
-  - osx
-
 cache:
   apt: true
   # We use three different cache directory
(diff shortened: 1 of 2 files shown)


vicramr:
TODO: for Mac, in `install.sh` the `deactivate` command causes a "command not found" message. Should only run this command on Linux.

lmcinnes:
Seems like it is almost working -- hopefully you can get over the last few hurdles. This will definitely be appreciated.

vicramr:
@lmcinnes There are 2 issues I'd like some input on:
1. It appears that for Mac, the coverage testing is taking significantly longer than on Linux (~35 min as opposed to ~20 min). This might be causing the build to time out (see addendum below). However, looking at the Travis build log, it appears that the tests did finish, and with a coverage of 93%. This is less than the coverage on Linux, which indicates that it might be valuable to run both coverage tests. Are you ok with just skipping the Mac coverage tests, or should we look into this further?
2. In my PR description I quoted a part of the Travis docs that basically says it's not worthwhile to cache downloaded files for the purpose of speeding up download time. Given this, are you ok with skipping the check for `miniconda.sh`?

Feel free to take your time in responding. There's no rush.

**Addendum:** I'm not actually sure why the build failed. The Travis docs [here](https://docs.travis-ci.com/user/customizing-the-build#build-timeouts) say that build timeouts occur when a single build takes over 50 minutes, or it produces no log output for over 10 minutes. But the former isn't true (it ran for 38 min) and I don't think the latter is either, because it *finished* the coverage tests (which is the only step that would even take 10+ minutes), started running `source ci_scripts/success.sh`, and then silently failed there.

Here's what I think happened: because we're `source`-ing `success.sh` and it has `set -e` at the beginning, if any of the commands in the script fails then the whole build will fail. This scenario is mentioned [here](https://docs.travis-ci.com/user/common-build-problems/#my-build-fails-unexpectedly) in the Travis docs. However, I'm not sure what command in the script could have failed.

lmcinnes:
Looks good now. Thanks for this!

PR #255, opened 2019-06-19
 Break out inner loop of optimize_layout so joblib can parallelize it

tomwhite:
If no random seed is set then `optimize_layout` will use all cores on the machine, otherwise it will use a single core for reprodcibility.

This supersedes #231. See previous discussion there.


lmcinnes:
This is looking very promising -- sorry for being slow to get to things lately. I'll try to give this a proper review soon. Please bear with me.

sleighsoft:
@tomwhite Can this PR be closed? If I remember correctly you had a new one that got merged?

PR #258, opened 2019-06-21
Add check for sparse matrix in fit

znwilkins:
Without this check, calling `len` on sparse matrices is ambiguous, and an error is thrown:
```
  File "../pythonEnv/lib/python3.6/site-packages/umap/umap_.py", line 1641, in fit_transform
    self.fit(X, y)
  File "../pythonEnv/lib/python3.6/site-packages/umap/umap_.py", line 1496, in fit
    if len(X) != len(y):
  File "../pythonEnv/lib/python3.6/site-packages/scipy/sparse/base.py", line 296, in __len__
    raise TypeError("sparse matrix length is ambiguous; use getnnz()"
```

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -1509,10 +1509,11 @@ class UMAP(BaseEstimator):
                 # )
 
         if y is not None:
-            if len(X) != len(y):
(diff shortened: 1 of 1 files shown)


lmcinnes:
Good catch! Thank you.

Issue #224, opened 2019-04-25
Is validation check for integers too narrow?


usul83:
I'm trying to incorporate UMAP into a pipeline which sets my n_component parameter to a numpy int64 at some point. This leads to a failed validation check (_validate_parameters(), line 1309 of umap_.py), since this is not an instance of a Python int. 

I suggest either making the check more broad or just converting to an int?

sleighsoft:
Will be resolved in 0.4

PR #269, opened 2019-07-18
Integer fix

usul83:
Following up on https://github.com/lmcinnes/umap/issues/224

The change means the validation code will first try to convert to a python int before raising an error if unable to. 

This is useful because other packages in a pipeline (e.g. scikit-optimize) can convert to numpy int64, which unfortunately is not an instance of a python int.


lmcinnes:
It seems like there is an issue since strings with integers in them can now pass through fine. While this might be fine, it is also possible that this is undesirable. Can you consider either amending the checks, or adjust the test cases appropriately to fit with the desired behaviour. Open to suggestions as to what the right behaviour here is.

sleighsoft:
@usul83 your contribution is very welcome but could you please update your PR to include the changes that lmcinnes suggested?

lmcinnes:
I changed the branch it is merging into, since 0.4 will hopefully be coming out soon, and it going to be explicitly python 3 so we can then skip some of those test failures. Hopefully this doesn't result in too many conflicts.

usul83:
I'm not sure why one of those Travis tests is still failing, but I've changed as per your concerns, it now fails if a string or float is passed, but attempts to convert anything else to a Python int.

lmcinnes:
The last Travis error is the formatting check test. Ideally all the code gets formatted with [``black``](https://black.readthedocs.io/en/stable/); currently it is complaining because ``umap/umap_.py`` would be reformatted by black. I would look at running ``black`` over the code and pushing that up to your branch. That should resolve the last issue.

lmcinnes:
Glad that worked! Thanks for this. Merging now.

PR #270, opened 2019-08-05
Added Jupiter notebook with Digits tutorial

zapaishchykova:
https://umap-learn.readthedocs.io/en/latest/basic_usage.html#digits-data


lmcinnes:
Hi, thanks for the PR. It would perhaps be beneficial if the notebooks were in the ``notebooks`` directory rather than at the root level. I'm also on vacation right now so I haven't had time to check the notebooks themselves too carefully.

sleighsoft:
Tried running it but it throws an error for me at 
```python
digits_df = pd.DataFrame(digits.data[:,:10])
digits_df['digit'] = pd.Series(digits.target).map(lambda x: 'Digit {}'.format(x))

sns.pairplot(digits_df, hue='digit', palette='Spectral');
```
_LinAlgError: singular matrix_

@zapaishchykova Can you try it on a fresh install of UMAP? And maybe it is better to fix the versions in the notebook. You could add `!pip install scikit=...` commands into the notebook.

PR #278, opened 2019-08-15
Use pynndescent dependency if installed, otherwise use nn descent in map (0.4dev version)

tomwhite:


lmcinnes:
I'll fix the travis error later -- it seems unrelated.

PR #279, opened 2019-08-16
Add a fast way to compute knn_indices 

sleighsoft:
Tested on a `[50k, 50k]` distance matrix computed on 50k `np.random.normal` points where this runs in 6.8s-7.0s on a 120 core machine whereas the old code takes 245s.

### DIFF
--- a/umap/utils.py
+++ b/umap/utils.py
@@ -25,9 +25,11 @@ def fast_knn_indices(X, n_neighbors):
     knn_indices: array of shape (n_samples, n_neighbors)
         The indices on the ``n_neighbors`` closest points in the dataset.
     """
-    knn_indices = np.empty((X.shape[0], n_neighbors), dtype=np.int32)
+    knn_indices = np.empty(
+        (X.shape[0], n_neighbors), dtype=np.int32
+    )
     for row in numba.prange(X.shape[0]):
-        v = X[row].argsort(kind='quicksort')
+        v = X[row].argsort(kind="quicksort")
         v = v[:n_neighbors]
         knn_indices[row] = v
     return knn_indices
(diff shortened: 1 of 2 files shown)


sleighsoft:
Fixed PEP 8 issues

umap/utils.py
```diff
@@ -8,6 +8,31 @@
...
+    n_neighbors: int
+        The number of nearest neighbors to compute for each sample in ``X``.
+
+    Returns
+    -------
+    knn_indices: array of shape (n_samples, n_neighbors)
+        The indices on the ``n_neighbors`` closest points in the dataset.
+    """
+    knn_indices = np.empty((X.shape[0], n_neighbors), dtype=np.int32)
+    for row in numba.prange(i.shape[0]):
```
lmcinnes: I believe you want ``X.shape[0]`` here.

lmcinnes:
Looks like a good plan -- thanks for this. Tests are currently failing, but it looks like a minor fix will get things working.

sleighsoft:
I am unsure why one test fails now.

lmcinnes:
There is a test for conformance to formatting with black. The easiest solution is to just run [black](https://github.com/psf/black) on the files you changed and commit the results. That should fix the remaining failure.

sleighsoft:
Ah, it probably is because you ran `black` after my fork :/

sleighsoft:
@lmcinnes Have you ever run black on umap_.py or utils.py?
If I do that it will change a lot of line I have not touched yet.

lmcinnes:
I believe it has happened, yes. Try it and see how it goes.

sleighsoft:
As you wish :D

sleighsoft:
Sigh, the error couldn't be less descriptive...

lmcinnes:
I'm honestly at a loss as well. I'll try to look a little deeper when I get some time.

sleighsoft:
Btw, if you find places where you are currently relying on argsort, searchsorted and sort you might want  to use some of my snippets here: https://github.com/sleighsoft/useful-python/blob/master/numba.py

lmcinnes:
I played with cache a while ago, and it had some issues at the time. I expect those are all gone by now, but I haven't had the time to go back and retry everything.

As for the options on argsort etc. -- I will definitely have a look, and would be happy to see pull requests using those tricks, as long as they are easy enough to read.

lmcinnes:
So that test seems to just be broken. I'm just going to merge this. Thanks for your work on it, and sorry for the hassles getting it in.

sleighsoft:
Didn't you do it the wrong way around now? `
Merge branch 'master' into patch-1 ` ?

sleighsoft:
Just want to make sure that you know that this is not yet merged with master

lmcinnes:
Yes, I was aware -- I had to resolve some conflicts and was waiting for the tests -- to make sure it was still only the one that failed after my conflict resolution efforts.

PR #285, opened 2019-08-28
Fix NumbaPerformanceWarning by removing `parallel=True` usages that were having no effect

tomwhite:
The latest Numba (0.45.1) is giving warnings like

```
umap/tests/test_umap.py::test_nn_search
  /Users/tom/workspace/umap/venv/lib/python3.7/site-packages/numba/compiler.py:602: NumbaPerformanceWarning: 
  The keyword argument 'parallel=True' was specified but no transformation for parallel execution was possible.
  
  To find out why, try turning on parallel diagnostics, see http://numba.pydata.org/numba-doc/latest/user/parallel.html#diagnostics for help.
  
  File "umap/nndescent.py", line 135:
      @numba.njit(parallel=True)
      def init_from_tree(tree, data, query_points, heap, rng_state):
      ^
  
    self.func_ir.loc))
```

This change fixes the warnings.

### DIFF
--- a/umap/nndescent.py
+++ b/umap/nndescent.py
@@ -43,7 +43,7 @@ def make_nn_descent(dist, dist_args):
     specialised to the given metric.
     """
 
-    @numba.njit(parallel=True)
+    @numba.njit()
     def nn_descent(
         data,
         n_neighbors,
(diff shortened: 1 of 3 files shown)


lmcinnes:
Thanks!

PR #286, opened 2019-08-28
Make new arguments to simplicial_set_embedding optional for backwards…

tomwhite:
… compatibility with 0.3.

With this change a;; scanpy master tests pass with umap 0.4dev. See https://github.com/theislab/scanpy/issues/779


gokceneraslan:
Thanks a lot @tomwhite!

PR #287, opened 2019-08-29
Fix more NumbaWarnings

tomwhite:


lmcinnes:
Thanks, I thought I had caught most of these.

PR #291, opened 2019-09-06
Handling for large numbers of duplicate samples

jc-healy:
I've added functionality to unique the rows of our data before embedding.  This ensures that identical data points are embedded to exactly the same location by umap.  I'm not copying the data but instead creating an index of unique rows and using that to access the data from then on.  The core of this depends on np.unique for the dense case and a new fairly simple function in umap.utils called csr_unique().

This behaviour can be activated via umap.UMAP(unique=True).   

Lastly, I've added a few unit tests to test_umap.py.


lmcinnes:
One error is that black reformatting is needed. The other seems to be an issue with fast_knn_indices, which apparently still has some issues compiling under numba on some platforms.

PR #314, opened 2019-10-24
Fix a typo

m-ueno:
Fix a typo.
`distances.pairwise_special_metric()` accepts keyword parameter `metric`, not `dmetric`.
This conditional branch looks not executed during CI.


lmcinnes:
Thanks; good catch. I really need to improve the test coverage in 0.4

PR #323, opened 2019-12-05
Set zip_safe=False to workaround numba cache issue

candalfigomoro:
We can get the `RuntimeError: cannot cache function 'rdist': no locator available` error when we try to use numba caching from a python egg package (e.g. if we installed umap through `python setup.py install`).

As a workaround, set the `zip_safe` flag to `False` to avoid installing it as egg file when we use the `setup.py install` command.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,8 @@ configuration = {
     'cmdclass' : {},
     'test_suite' : 'nose.collector',
     'tests_require' : ['nose'],
-    'data_files' : ()
+    'data_files' : (),
+    'zip_safe' : False
     }
 
 setup(**configuration)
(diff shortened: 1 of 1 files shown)


sleighsoft:
First of all, thank you for your PR.

Could it be that you commited different line endings than what UMAP uses?
This would explain why your PR has many changed lines while you actually only added a single line.
Please verify and if that is the case update your PR.

candalfigomoro:
Hi @sleighsoft,
actually many lines changed because I removed the whitespace before the colons in order to pass the PEP8 checks.

sleighsoft:
I see. Considering that UMAP is not a very large package I think there is no downside to changing this.
Looks good to me.

lmcinnes:
Thanks all!

Issue #324, opened 2019-12-10
"No conversion from array(float64, 2d, A) to array(float32, 2d, C) for 'knn_dists'" when calling .transform() in 0.4.0rc1


aparrish:
After installing `umap-learn` from (what looked to me like) the dev branch like so:

```
pip install https://github.com/lmcinnes/umap/archive/0.4dev.zip
```

I get a weird error apparently about the data type of some internal value. Example code with stack trace included below.

This code works fine for me (in the same environment) with 0.3.10 installed instead. Strangely, it also works when I install the `0.4dev_fix#306` branch (with the pip/github zip method mentioned above). But I couldn't decipher from looking at the compare interface if `0.4dev_fix#306` is actually newer than `0.4dev`?

```
>>> import numpy as np
>>> import umap
>>> X = np.random.uniform(size=(10000, 100))
>>> mapper = umap.UMAP().fit(X)
>>> mapper.transform(X[:10])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/allison/tmp/umap-testing/env3/lib/python3.7/site-packages/umap/umap_.py", line 1972, in transform
    rows, cols, vals = compute_membership_strengths(indices, dists, sigmas, rhos)
  File "/Users/allison/tmp/umap-testing/env3/lib/python3.7/site-packages/numba/dispatcher.py", line 401, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/Users/allison/tmp/umap-testing/env3/lib/python3.7/site-packages/numba/dispatcher.py", line 344, in error_rewrite
    reraise(type(e), e, None)
  File "/Users/allison/tmp/umap-testing/env3/lib/python3.7/site-packages/numba/six.py", line 668, in reraise
    raise value.with_traceback(tb)
numba.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No conversion from array(float64, 2d, A) to array(float32, 2d, C) for 'knn_dists', defined at None

File "env3/lib/python3.7/site-packages/umap/umap_.py", line 440:
def compute_membership_strengths(knn_indices, knn_dists, sigmas, rhos):
    <source elided>
    """
    n_samples = knn_indices.shape[0]
    ^
```

This is in a fresh Python 3.7.0 venv (MacOS 10.14.6). My `pip list`:

```
Package      Version 
------------ --------
joblib       0.14.1  
llvmlite     0.30.0  
numba        0.46.0  
numpy        1.17.4  
[...]

PR #327, opened 2019-12-12
convert dists earlier in transform (fix for #324)

aparrish:
Proposed fix for #324. Upstream version passes a copy of `dists` converted to numpy type `float32` in the call to `smooth_knn_dists` but used the original value subsequently in `compute_membership_strengths`, which is why I think I was continuing to get the error in the referenced issue. This fix converts `dists` upfront to `float32`, which seems to fix it!

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -1963,9 +1963,11 @@ class UMAP(BaseEstimator):
             indices = indices[:, : self._n_neighbors]
             dists = dists[:, : self._n_neighbors]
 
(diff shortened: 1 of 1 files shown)


lmcinnes:
Good catch; thanks! Greatly appreciated.

jan-engelmann:
thanks for the fix! doesn't seem to be released yet is it?
I am using umap==0.5.4 installed with pip

lmcinnes:
Ah; I didn't even look -- it got merged into a different branch. I'll try to fix this soon.

PR #329, opened 2019-12-17
Correct handling of indices in tails of sparse_sum

archmageirvine:
This pull request fixes what appears to be a bug in the `sparse_sum` function in the `sparse.py` class.  It appears that the piece of code dealing with the tails of the vectors is incorrect.  In particular, the lines like `result_ind[nnz] = i1` should be `result_ind[nnz] = ind1[i1]` to ensure the original vector indices are preserved.


lmcinnes:
Thanks -- a silly error, but a good catch (since this is buried pretty deep in things). Greatly appreciated.

PR #332, opened 2020-01-01
docstring in smooth_knn_dist changed

hyungkwonko:
Return values are changed from knn_dist, nn_dist to result, rho.

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -96,7 +96,7 @@ def smooth_knn_dist(
     result: array of shape(n_samples)
         The normalization factor derived from the metric tensor approximation.
 
-    rhos: array of shape(n_samples)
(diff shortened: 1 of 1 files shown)


umap/umap_.py
```diff
@@ -93,11 +93,11 @@ def smooth_knn_dist(
 
     Returns
     -------
-    knn_dist: array of shape (n_samples,)
-        The distance to kth nearest neighbor, as suitably approximated.
+    result: array of shape(n_samples)
```
sleighsoft: I think this should rather be called `sigmas` to be in-line with https://github.com/lmcinnes/umap/blob/f3f2dac5517d065bf799c351da0d3f3d2203a02b/umap/umap_.py#L326

PR #335, opened 2020-01-05
Add legends to plots when labels are provided; minor bug fixes for diagnostic_type='all' plots

greg-campbell:
1. Allow automatic plot legends when passing labels param; can be turned off with show_legend=False
2. Show all diagnostic plots when diagnostic_type='all' (was missing neighborhood plot)
3. Fix layout when plotting diagnostic_type='all'. Before, it showed an empty subplot at the top.

### DIFF
--- a/umap/plot.py
+++ b/umap/plot.py
@@ -231,7 +231,7 @@ def _datashade_points(
     background="white",
     width=800,
     height=800,
-    show_legend=True
+    show_legend=True,
 ):
 
     """Use datashader to plot points"""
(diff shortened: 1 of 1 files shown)


lmcinnes:
Hi, this looks great. There are some errors, but I think they are all upstream of this. You may want to run the latest version of [black](https://black.readthedocs.io/en/stable/) over the code however to get the formatting right.

lmcinnes:
Thanks -- it looks like all the issues are now upstream, so I'll merge this. Thanks again for your work on this, and please feel free to submit more PRs like this anytime!

PR #339, opened 2020-01-13
Correcting some validation message errors in _validate_parameters.

greg-campbell:
The errors didn't match the conditions that were being checked.


lmcinnes:
Thank you -- good catches.

PR #341, opened 2020-01-15
Default show_legend to False when calling umap.plot.connectivity with show_points = True.

### LINKED ISSUE #340: show_legend missing in umap.plot.connectivity
When calling umap.plot.connectivity(mapper, show_points=True) the function tries to call  _datashade_points and pass the parameter show_legend but this variable is not defined.

This is in the version installed with:
 pip install git+https://github.com/lmcinnes/umap.git@0.4dev

greg-campbell:
Fixes #340

### DIFF
--- a/umap/plot.py
+++ b/umap/plot.py
@@ -801,7 +801,7 @@ def connectivity(
             None,
             width,
             height,
-            show_legend,
+            False,
         )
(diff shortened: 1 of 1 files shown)


lmcinnes:
Thanks.

Issue #337, opened 2020-01-11
Plotting a subset of points in umap.plot


gclen:
Creating this issue to go along with a pull request. 

It's often useful to be able to select a subset of points to plot when exploring your data (e.g. ignoring outliers). It should be easy to select the subset (just an array of booleans) with the same number of points as the input data.

```python
import pandas as pd
#import numpy as np
import umap
import umap.plot
from sklearn.datasets import load_digits
#Some plotting libraries
import matplotlib.pyplot as plt
%matplotlib notebook
from bokeh.plotting import show, save, output_notebook, output_file
from bokeh.resources import INLINE 
output_notebook(resources=INLINE)

digits = load_digits()

embedding = umap.UMAP().fit(digits.data)

hover_df = pd.DataFrame(digits.target, columns=['target'])
# Select digits less than 5
# This is just an array of True/False
subset = hover_df['target'] < 5

f = umap.plot.interactive(embedding, labels=hover_df['target'], hover_data=hover_df, subset_points=subset)
show(f)
```

PR #342, opened 2020-01-17
Add subset_points argument to points plotting method

gclen:
Do the same thing as https://github.com/lmcinnes/umap/issues/337 but for the non-interactive method.


lmcinnes:
Thanks; looks good.

PR #347, opened 2020-01-24
Move Installation closer to the beginning

sleighsoft:
Many people confuse the package to be called `umap` instead of `umap-learn`. Maybe moving the installation instructions to the top prevents this issue.


lmcinnes:
Looks good. Thanks. I still need to fix a bunch of errors in the build, but I'll try to deal with that when I get some time.

PR #352, opened 2020-01-28
plot.py: Use `color_key` when datashading labels

wmayner:
Pass `color_key` to datashader when labels are provided. When `ds.count_cat` is used, the colors of the categories can be specified with `color_key`.

This allows the user to specify colors when plotting large data with labels; currently, colors cannot be specified, and if there are more than 22 labels this codepath throws an error.

Also, remove the `vdim` on the `hv.Points` plot since it is not used.


lmcinnes:
Thank you!

PR #369, opened 2020-03-03
custom distance metrics for transform & inverse transform, plus some other bug fixes

### LINKED ISSUE #366: fit/transform/inverse_transform broken for custom distance metrics
hey @lmcinnes, thanks for maintaining such an awesome package.  Looks like 0.4 is getting close to release -- super exciting!  

I'm running into a couple issues with using custom distance functions (using the repo, master branch, most recent commit lmcinnes/umap@1a3892610643ca0ba72a0c559488f47e43bbd7cf):
1. if a callable `output_metric` is passed, [`simplicial_set_embedding` calls `optimize_layout_generic`](https://github.com/lmcinnes/umap/blob/1a3892610643ca0ba72a0c559488f47e43bbd7cf/umap/umap_.py#L1103), which fails [here](https://github.com/lmcinnes/umap/blob/1a3892610643ca0ba72a0c559488f47e43bbd7cf/umap/layouts.py#L341) with a numba `TypingError`, because it expects to be using the gradient-returning version of a built-in named metric, which would return a tuple.  This would also affect the `transform` method [here](https://github.com/lmcinnes/umap/blob/1a3892610643ca0ba72a0c559488f47e43bbd7cf/umap/umap_.py#L2074) for the same reason.
2. Similarly, if a callable `metric` is passed, the `inverse_transform` method won't work because `optimize_layout_inverse` fails [here](https://github.com/lmcinnes/umap/blob/1a3892610643ca0ba72a0c559488f47e43bbd7cf/umap/layouts.py#L505) for the same reason.

I'm planning to try to fix this on my own fork anyway, so I'd be happy to submit a PR when I'm done if you're interested! There are a couple different ways I think it could be addressed, depending on what you'd prefer, but each is a bit tricky and has its downsides.

Oh, also, it looks like the fixes I submitted in #330 were overwritten by a subsequent pull request.  If you want, I could re-add those changes.

Thanks again!

### LINKED ISSUE #367: fit method hangs after recent updates
I can't figure out why this is happening, but after some recent changes to `master` branch, the `.fit` method hangs to the point I have to kill python to avoid it crashing my computer.

The problem doesn't exist as of the most recent 0.4 release candidate (fc59aa78f14b10e5eafc19b18fc135de1c5382ea), but does with the most recent version of `master` (1a3892610643ca0ba72a0c559488f47e43bbd7cf).

It appears to be coming from a super slow list comprehension `scipy` runs internally to transpose a sparse matrix, called from `umap` [here](https://github.com/lmcinnes/umap/blob/1a3892610643ca0ba72a0c559488f47e43bbd7cf/umap/umap_.py#L1696).  It doesn't seem like a bug in `scipy` though, since my version is the same for both the working and broken examples (`1.2.1`).

<details>
<summary>Working version:</summary>

```
import numpy as np
from time import time
from umap import UMAP

to_reduce = np.random.uniform(0, 1, (5000, 25))
reducer = UMAP(random_state=0, metric='euclidean', verbose=True)

t1 = time()
reducer.fit(to_reduce)
t2 = time()
print(t2 - t1)

--------------------------------------------------------
UMAP(a=None, angular_rp_forest=False, b=None,
     force_approximation_algorithm=False, init='spectral', learning_rate=1.0,
     local_connectivity=1.0, low_memory=False, metric='euclidean',
     metric_kwds=None, min_dist=0.1, n_components=2, n_epochs=None,
     n_neighbors=15, negative_sample_rate=5, output_metric='euclidean',
     output_metric_kwds=None, random_state=0, repulsion_strength=1.0,
     set_op_mix_ratio=1.0, spread=1.0, target_metric='categorical',
     target_metric_kwds=None, target_n_neighbors=-1, target_weight=0.5,
     transform_queue_size=4.0, transform_seed=42, unique=False, verbose=True)
Construct fuzzy simplicial set
Mon Mar  2 14:57:26 2020 Finding Nearest Neighbors
Mon Mar  2 14:57:26 2020 Building RP forest with 9 trees
Mon Mar  2 14:57:26 2020 NN descent for 12 iterations
	 0  /  12
	 1  /  12
	 2  /  12
	 3  /  12
	 4  /  12
	 5  /  12
	 6  /  12
	 7  /  12
	 8  /  12
	 9  /  12
	 10  /  12
	 11  /  12
Mon Mar  2 14:57:33 2020 Finished Nearest Neighbor Search
Mon Mar  2 14:57:35 2020 Construct embedding
	completed  0  /  500 epochs
	completed  50  /  500 epochs
	completed  100  /  500 epochs
	completed  150  /  500 epochs
	completed  200  /  500 epochs
	completed  250  /  500 epochs
	completed  300  /  500 epochs
	completed  350  /  500 epochs
	completed  400  /  500 epochs
	completed  450  /  500 epochs
Mon Mar  2 14:57:47 2020 Finished embedding
21.280228853225708
```
</details>

<details>
<summary>Broken version:</summary>

```
import numpy as np
from time import time
from umap import UMAP

to_reduce = np.random.uniform(0, 1, (5000, 25))
reducer = UMAP(random_state=0, metric='euclidean', verbose=True)

t1 = time()
reducer.fit(to_reduce)
t2 = time()
print(t2 - t1)

--------------------------------------------------------
UMAP(a=None, angular_rp_forest=False, b=None,
     force_approximation_algorithm=False, init='spectral', learning_rate=1.0,
     local_connectivity=1.0, low_memory=False, metric='euclidean',
     metric_kwds=None, min_dist=0.1, n_components=2, n_epochs=None,
     n_neighbors=15, negative_sample_rate=5, output_metric='euclidean',
     output_metric_kwds=None, random_state=0, repulsion_strength=1.0,
     set_op_mix_ratio=1.0, spread=1.0, target_metric='categorical',
     target_metric_kwds=None, target_n_neighbors=-1, target_weight=0.5,
     transform_queue_size=4.0, transform_seed=42, unique=False, verbose=True)
Construct fuzzy simplicial set
Mon Mar  2 14:52:09 2020 Finding Nearest Neighbors
Mon Mar  2 14:52:09 2020 Building RP forest with 9 trees
Mon Mar  2 14:52:10 2020 NN descent for 12 iterations
	 0  /  12
	 1  /  12
	 2  /  12
	 3  /  12
	 4  /  12
	 5  /  12
	 6  /  12
	 7  /  12
	 8  /  12
	 9  /  12
	 10  /  12
	 11  /  12
Mon Mar  2 14:52:17 2020 Finished Nearest Neighbor Search
---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
<ipython-input-3-e2b9ca9bf5a6> in <module>()
      1 t1 = time()
----> 2 reducer.fit(to_reduce)
      3 t2 = time()
      4 print(t2 - t1)

/opt/conda/lib/python3.6/site-packages/umap/umap_.py in fit(self, X, y)
   1694                 self._search_graph.data = _data
   1695                 self._search_graph = self._search_graph.maximum(
-> 1696                     self._search_graph.transpose()
   1697                 ).tocsr()
   1698 

/opt/conda/lib/python3.6/site-packages/scipy/sparse/lil.py in transpose(self, axes, copy)
    456 
    457     def transpose(self, axes=None, copy=False):
--> 458         return self.tocsr(copy=copy).transpose(axes=axes, copy=False).tolil(copy=False)
    459 
    460     transpose.__doc__ = spmatrix.transpose.__doc__

/opt/conda/lib/python3.6/site-packages/scipy/sparse/lil.py in tocsr(self, copy)
    473 
    474         indptr = np.cumsum([0] + lst, dtype=idx_dtype)
--> 475         indices = np.array([x for y in self.rows for x in y], dtype=idx_dtype)
    476         data = np.array([x for y in self.data for x in y], dtype=self.dtype)
    477 

/opt/conda/lib/python3.6/site-packages/scipy/sparse/lil.py in <listcomp>(.0)
    473 
    474         indptr = np.cumsum([0] + lst, dtype=idx_dtype)
--> 475         indices = np.array([x for y in self.rows for x in y], dtype=idx_dtype)
    476         data = np.array([x for y in self.data for x in y], dtype=self.dtype)
    477 

KeyboardInterrupt: 
```
</details>

Note: this only happens when the data isn't "small" and the distance matrix doesn't get pre-computed

paxtonfitzpatrick:
### This PR:
- allows custom distance metrics to be used for fitting, transforming, and inverse-transforming (fixes #366).  This is done by:
  - quickly checking a callable `metric` on the input data to see whether or not it returns a gradient in addition to distance (new `self._check_custom_metric` function)
  - if it does, wrapping it in a `numba.njit`'d function that returns only the distance and using that for `.fit`/`.transform`, while using the full function for `.transform`.
  - if it doesn't, warning upfront that `inverse_transform` will fail, and then explicitly failing if `inverse_transform` is called
  - checking a custom `output_metric` on some simulated `n_components`-dimensional data, and failing with some explanation why
- validates named sparse distance metrics upfront to fail earlier/faster and avoid dynamically updating `UMAP` attributes later
- fixes pre-computing distance matrix for small data using custom metrics that take kwargs
- fixes bug in `inverse_transform` when projecting more data than was used to fit model
- fixes bug in `self._search_graph` construction (fixes #367)
- consolidates a bunch of duplicate param checks and assignments into `self._validate_parameters` and `__init__` (doesn't add any new failure conditions, just fails quicker)
- removes no-longer-needed aliasing of `self.metric_kwargs` (`self._metric_kwargs`)
- removes some other no-longer-needed aliasing of attributes in `.transform` and `.inverse_transform`
- adds some new file types to `.gitignore` (+ some info)
- all changes to files under `docs` and `examples`, and in `umap/plot.py` are due to running `black`

### This PR _doesn't_:
- Implement these new features in the `DataFrameUMAP` class (which currently requires that metrics be named distance functions)
- alter the outward-facing API in any way (to the best of my knowledge)
- update docs to reflect new features (e.g., noting that callable `metric`s should return a tuple of (distance, gradient) if the user wants to be able to use the `inverse_transform` method)
- add any new tests for the new features (though they're largely covered by the existing ones)
- handle the case where a user: 1) has `pynndescent` installed and wants to use it, 2) tries to use `.fit` on a sparse matrix, and 3) passes a custom `metric` function takes some keyword arguments.  This is because `pynndescent` [doesn't accept callable metrics](https://github.com/lmcinnes/pynndescent/blob/13cf8e0a9f60741e4c1847a5447648fc4c1cae8d/pynndescent/pynndescent_.py#L933-L948) for `pynndescent.sparse_nnd.nn_descent`.  Other than this specific combination of circumstances, everything seems to work fine as far as I can tell.

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -1380,14 +1380,8 @@ class UMAP(BaseEstimator):
         self.metric = metric
         self.output_metric = output_metric
         self.target_metric = target_metric
-        if metric_kwds is None:
-            self.metric_kwds = {}
-        else:
-            self.metric_kwds = metric_kwds
-        if output_metric_kwds is None:
-            self.output_metric_kwds = {}
-        else:
-            self.output_metric_kwds = output_metric_kwds
+        self.metric_kwds = metric_kwds
+        self.output_metric_kwds = output_metric_kwds
         self.n_epochs = n_epochs
         self.init = init
         self.n_components = n_components
@@ -1405,10 +1399,7 @@ class UMAP(BaseEstimator):
         self.transform_queue_size = transform_queue_size
         self.target_n_neighbors = target_n_neighbors
         self.target_metric = target_metric
-        if target_metric_kwds is None:
-            self.target_metric_kwds = {}
-        else:
-            self.target_metric_kwds = target_metric_kwds
+        self.target_metric_kwds = target_metric_kwds
         self.target_weight = target_weight
         self.transform_seed = transform_seed
         self.force_approximation_algorithm = force_approximation_algorithm
@@ -1463,6 +1454,18 @@ class UMAP(BaseEstimator):
             self.n_epochs <= 10 or not isinstance(self.n_epochs, int)
         ):
             raise ValueError("n_epochs must be a positive integer of at least 10")
+        if self.metric_kwds is None:
+            self._metric_kwds = {}
+        else:
+            self._metric_kwds = self.metric_kwds
+        if self.output_metric_kwds is None:
+            self._output_metric_kwds = {}
+        else:
+            self._output_metric_kwds = self.output_metric_kwds
+        if self.target_metric_kwds is None:
+            self._target_metric_kwds = {}
+        else:
+            self._target_metric_kwds = self.target_metric_kwds
         # check sparsity of data upfront to set proper _input_distance_func &
         # save repeated checks later on
         if scipy.sparse.isspmatrix_csr(self._raw_data):
@@ -1472,7 +1475,7 @@ class UMAP(BaseEstimator):
         # set input distance metric & inverse_transform distance metric
         if callable(self.metric):
             in_returns_grad = self._check_custom_metric(
-                self.metric, self.metric_kwds, self._raw_data
+                self.metric, self._metric_kwds, self._raw_data
             )
             if in_returns_grad:
                 _m = self.metric
@@ -1529,7 +1532,7 @@ class UMAP(BaseEstimator):
         # set ooutput distance metric
         if callable(self.output_metric):
             out_returns_grad = self._check_custom_metric(
-                self.output_metric, self.output_metric_kwds
+                self.output_metric, self._output_metric_kwds
             )
             if out_returns_grad:
                 self._output_distance_func = self.output_metric
@@ -1689,7 +1692,7 @@ class UMAP(BaseEstimator):
             try:
                 # sklearn pairwise_distances fails for callable metric on sparse data
                 _m = self.metric if self._sparse_data else self._input_distance_func
-                dmat = pairwise_distances(X[index], metric=_m, **self.metric_kwds)
+                dmat = pairwise_distances(X[index], metric=_m, **self._metric_kwds)
             except (ValueError, TypeError) as e:
                 # metric is numba.jit'd or not supported by sklearn,
                 # fallback to pairwise special
@@ -1697,20 +1700,20 @@ class UMAP(BaseEstimator):
                     dmat = dist.pairwise_special_metric(
                         X[index].toarray(),
                         metric=self._input_distance_func,
-                        kwds=self.metric_kwds,
+                        kwds=self._metric_kwds,
                     )
                 else:
                     dmat = dist.pairwise_special_metric(
                         X[index],
                         metric=self._input_distance_func,
-                        kwds=self.metric_kwds,
+                        kwds=self._metric_kwds,
                     )
             self.graph_, self._sigmas, self._rhos = fuzzy_simplicial_set(
                 dmat,
                 self._n_neighbors,
                 random_state,
                 "precomputed",
-                self.metric_kwds,
+                self._metric_kwds,
                 None,
                 None,
                 self.angular_rp_forest,
@@ -1731,7 +1734,7 @@ class UMAP(BaseEstimator):
                 X[index],
                 self._n_neighbors,
                 nn_metric,
-                self.metric_kwds,
+                self._metric_kwds,
                 self.angular_rp_forest,
                 random_state,
                 self.low_memory,
@@ -1744,7 +1747,7 @@ class UMAP(BaseEstimator):
                 self.n_neighbors,
                 random_state,
                 nn_metric,
-                self.metric_kwds,
+                self._metric_kwds,
                 self._knn_indices,
                 self._knn_dists,
                 self.angular_rp_forest,
@@ -1771,10 +1774,10 @@ class UMAP(BaseEstimator):
                     self._search_graph.transpose()
                 ).tocsr()
 
-                if (self.metric != "precomputed") and (len(self.metric_kwds) > 0):
+                if (self.metric != "precomputed") and (len(self._metric_kwds) > 0):
                     # Create a partial function for distances with arguments
                     _distance_func = self._input_distance_func
-                    _dist_args = tuple(self.metric_kwds.values())
+                    _dist_args = tuple(self._metric_kwds.values())
                     if self._sparse_data:
 
                         @numba.njit()
@@ -1843,13 +1846,13 @@ class UMAP(BaseEstimator):
                         ydmat = pairwise_distances(
                             y_[np.newaxis, :].T,
                             metric=self.target_metric,
-                            **self.target_metric_kwds
+                            **self._target_metric_kwds
                         )
                     except (TypeError, ValueError):
                         ydmat = dist.pairwise_special_metric(
                             y_[np.newaxis, :].T,
                             metric=self.target_metric,
-                            kwds=self.target_metric_kwds,
+                            kwds=self._target_metric_kwds,
                         )
 
                     target_graph, target_sigmas, target_rhos = fuzzy_simplicial_set(
@@ -1857,7 +1860,7 @@ class UMAP(BaseEstimator):
                         target_n_neighbors,
                         random_state,
                         "precomputed",
-                        self.target_metric_kwds,
+                        self._target_metric_kwds,
                         None,
                         None,
                         False,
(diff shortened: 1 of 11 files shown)


paxtonfitzpatrick:
@lmcinnes let me know if I should change anything or explain why I did something! 

Also, I've been using a custom distance metric that computes the symmetrized KL divergence between two distributions, and have a gradient calculation implemented for it as well.  I held off on adding it without your approval, but if you're interested, I'd love to add it as a precomputed metric for convenience.

lmcinnes:
Thanks so much for all of this! It looks amazing at first glance. I'll have to take a little while to dig through it all, but it mostly looks good to go.

I would be more than happy to have another named metric added, particularly if it is a useful one (and Jensen-Shannon, or similar, would certainly be useful). Please add it in!

umap/umap_.py
```diff
@@ -1387,16 +1376,18 @@ def __init__(
...
+        if metric_kwds is None:
+            self.metric_kwds = {}
         else:
-            self._output_metric_kwds = {}
-
+            self.metric_kwds = metric_kwds
+        if output_metric_kwds is None:
+            self.output_metric_kwds = {}
+        else:
+            self.output_metric_kwds = output_metric_kwds
```
lmcinnes: The sklearn API says that the named attributes should *always* match the parameters that were passed in; that was the reason for the odd indirection with private/underscored attributes holding the values that were actually going to be used in computation (paired with the fact that passing (empty) dicts as default parameters is bad). I would like to try and keep that if possible, even if it is a little awkward. It shouldn't be too tricky to thread that through here...

umap/umap_.py
```diff
@@ -1652,13 +1720,18 @@ def fit(self, X, y=None):
                 self.verbose,
             )
         else:
-            self._small_data = False
             # Standard case
+            self._small_data = False
+            # pynndescent doesn't accept callable functions for sparse data
+            if _HAVE_PYNNDESCENT and self._sparse_data and isinstance(self.metric, str):
```
lmcinnes: Since pynndescent is the "future" for all the NN computation that's where a lot of work is going, including efforts to generally speed things up for all the metrics that can be sped up. I am unlikely to port those improvements back to umap, so it would be good if we could pass the string version of the metric through to pynndescent if at all possible and let it choose the best option. That means you can simplify this condition to just if _HAVE_PYNNDESCENT:

umap/umap_.py
paxtonfitzpatrick: Ah, sorry about that, wasn't editing with those guidelines in mind.  Happy to add it back in and make sure I didn't violate them elsewhere.  How strictly do you want the syntax to adhere to them (e.g., it looks like `self._raw_data` probably "should" be `self.raw_data_`)?

umap/umap_.py
paxtonfitzpatrick: Sure 👍 Though I think it'd still have to be `if _HAVE_PYNNDESCENT and isinstance(self.metric, str)`, in order to account for a user passing a callable to `self.metric`, right?  
I chose to preferentially pass the callable version of the metric in case there were any named metrics in `umap` that didn't also exist in `pynndescent`.  Is that a condition worth worrying about?

umap/umap_.py
paxtonfitzpatrick: fixed

umap/umap_.py
paxtonfitzpatrick: tried to do this in a reasonably safe way --
- when trying to import `pynndescent`, also import the names of distance functions defined in `pynndescent.distances` and `pynndescent.sparse`
- if `_HAVE_PYNNDESCENT`, prefer passing the string version of the metric as long as `pynndescent` also defines it (for the appropriate data type)

That should enable passing the string version more often as UMAP's additional metrics are ported over to `pynndescent`, while also remaining functional if new metrics are added to `umap`

umap/umap_.py
lmcinnes: In theory if the attribute name ends with an underscore it is a property of the model generated by the fitting process that should be meaningful (i.e. is public); I'm not convinced, for instance, that the raw data should be public, so I use the prefix underscore for that and other similar things.

umap/umap_.py
lmcinnes: Yes -- that sounds like an excellent compromise. Thanks.

lmcinnes:
Looks good to me -- thanks for all your work on this!

Issue #361, opened 2020-02-18
subset_points not working for interactive plots


IA-Cardiologia-husa:
Hi!

First of all, thank you so much for this amazing library.

I have noticed a little issue while plotting a subset of points with interactive plots.
In the function `interactive` of `umap/plot.py`` [LINE 1268]
instead of:
"""
points = umap_object.embedding_
    if subset_points:
        points = points[subset_points]
"""
I think it would be better to just use:

as you do in the points function:
"""
if subset_points is not None:
        if len(subset_points) != points.shape[0]:
            raise ValueError(
                "Size of subset points ({}) does not match number of input points ({})".format(
                    len(subset_points), points.shape[0]
                )
            )

        data = data[subset_points]
        if hover_data is not None:
            hover_data = hover_data[subset_points]
"""

Thanks

lmcinnes:
I agree -- I'll try to get that added shortly; or you are welcome to submit a pull request if you would like it sooner.

PR #370, opened 2020-03-04
Make subset_plots more robust

gclen:
Fix for https://github.com/lmcinnes/umap/issues/361


lmcinnes:
I think you just need to run a recent version of black over plot.py and then this is good to go. Thanks Graham!

gclen:
Done!

sleighsoft:
Looks good to me :)

PR #373, opened 2020-03-06
Add document embedding example to docs

gclen:
Convert [document embedding notebook](https://github.com/lmcinnes/umap/blob/master/notebooks/Document%20embedding%20using%20UMAP.ipynb) to rst to be used in readthedocs


lmcinnes:
Thanks! Always glad to have more documentation, especially for use cases.

PR #382, opened 2020-03-27
Moved the metadata into setup.cfg

KOLANICH:


lmcinnes:
Thanks! I admit that I am not very good at the whole packaging thing, so I'm happy to just get stuff to work at all usually. I definitely appreciate work that can make this more efficient and maintainable (and up to date with modern methods for doing these things).

sleighsoft:
I think the dependency `setuptools_scm` uses git to craft version strings. I haven't looked into it too much though.

PR #391, opened 2020-04-05
Updating plot.py import error message to remove seaborn and add colorcet

### LINKED ISSUE #390: Is seaborn requred?
https://github.com/lmcinnes/umap/blob/499f8350b3374767f41fcd49eac13b982c9329ac/umap/plot.py#L32

I'm not seeing the explicit import for seaborn. Is this involved in any of the other packages?

anzelpwj:
Fixes #390 .

I've updated the error messge for plot.py as well as changing the extra requirements in setup.py.

With that said, the code in `examples` *do* use seaborn. If it makes sense to include `seaborn` in the setup file for these, I can add it back.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ configuration = {
         "tbb >= 2019.0",
     ],
     "extras_require": {
-        "plot": ["matplotlib", "datashader", "bokeh", "holoviews", "seaborn"],
+        "plot": ["matplotlib", "datashader", "bokeh", "holoviews", "colorcet"],
         "performance": ["pynndescent >= 0.4"],
     },
     "ext_modules": [],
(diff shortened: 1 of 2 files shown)


anzelpwj:
Found an error in CI, but seems unrelated to the work I did:

```
numba.errors.LoweringError: Failed in nopython mode pipeline (step: nopython mode backend)
Operands must be the same type, got (i64, i32)
File "../../../Users/travis/build/lmcinnes/umap/umap/utils.py", line 160:
def make_heap(n_points, size):
    <source elided>
    result = np.zeros(
        (np.int32(3), np.int32(n_points), np.int32(size)), dtype=np.float64
```

lmcinnes:
Yeah -- that's an odd ongoing error that only seems to occur in some off corner cass of the CI and I haven't been able to reproduce it very well.

Thanks for this -- and for noticing that colorcet is needed. Greatly appreciated!

PR #396, opened 2020-04-07
Patch init import to allow for local dev

leriomaggio:
This is very simple PR, suggesting two changes to to ease the local dev without the need to actually install `umap-learn` in a local env. 

The PR brings two changes:
- modify `umap.__init__.py` to allow working on a dev local version (more on this below)

- including in the `.gitignore` file the `other.xml` idea file - also non-stylistic and so to be ignored.

### The Issue when importing `umap` in a local dev environment

When importing `umap`, the `__version__` variable is automatically derived from `pkg_resources`.
This does not work without installing the package. 

### The (simple and silly) Solution

Catching the exception and manually setting up the version. 
As far as I can see, this won't interfere in any case with normally setup package in a virtual environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ venv
 .idea/vcs.xml
 .idea/workspace.xml
 .idea/dictionaries
+.idea/other.xml
 
 # Mac Finder layout
 .DS_Store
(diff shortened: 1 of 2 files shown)


lmcinnes:
Thanks! I admit that I don't have much expertise with these things, so if this helps others develop with UMAP all the better.

PR #407, opened 2020-04-16
fixed mnist.loader import statement

sg-s:
In the documentation, on the supervised clustering page, the example has the following statement to import the fashion MNIST dataset:

```
from mnist import MNIST
```

I believe this is an error, and it should be

```
from mnist.loader import MNIST
```

This is the only change.


sg-s:
This build failure seems unrelated to the change I made.

lmcinnes:
Thanks!

Issue #424, opened 2020-05-07
Passing sparse distance matrices or KNNs directly


Rocketknight1:
Hey, I'm trying to do clustering on approx 300k samples using the [Word Mover's Distance](http://proceedings.mlr.press/v37/kusnerb15.pdf) as a metric. This means that the input data does not take the form of vectors, and all-pairs distance comparisons are very expensive to compute. However, I can efficiently get all-pairs (approximate) nearest neighbours using methods like the 'centroid distance' from the paper and populate a sparse distance matrix with the distance values from all points to their nearest neighbours.

I'd really like to try passing those sparse distances to UMAP, but I get the following error when I try:

```
TypingError: Failed in nopython mode pipeline (step: nopython frontend)
non-precise type pyobject
[1] During: typing of argument at /home/matt/miniconda3/envs/story-clustering/lib/python3.7/site-packages/umap_learn-0.4.2-py3.7.egg/umap/utils.py (29)

File "../../../miniconda3/envs/story-clustering/lib/python3.7/site-packages/umap_learn-0.4.2-py3.7.egg/umap/utils.py", line 29:
def fast_knn_indices(X, n_neighbors):
    <source elided>
    """
    knn_indices = np.empty((X.shape[0], n_neighbors), dtype=np.int32)
    ^

This error may have been caused by the following argument(s):
- argument 0: cannot determine Numba type of <class 'scipy.sparse.csr.csr_matrix'>
```

Is there any way I can get around this? I can compute nearest neighbours for free just by sorting rows of the sparse distance matrix, but I don't understand the UMAP code well enough to hack my precomputed NNs in. The distance matrix is also much too big to convert to dense form - a 300,000 x 300,000 matrix of float32 values would take ~340GB of memory, so I'm kind of stuck!

Here's my very simple code snippet that triggers this error:

```
data = scipy.sparse.load_npz('sparse_distances.npz').tocsr()
fit = umap.UMAP(metric='precomputed')
embedding = fit.fit_transform(data)
```

lmcinnes:
I believe the issue is that precomputed sparse matrices are not currently supported. I'm not sure if there is an easy workaround at this time. In practice you can probably reach into the innards of UMAP and force it to work with what you have, but that's a little non-trivial. If you are interested in this option we can try to work out a reasonable set of changes you could make to make it work.

[...]

PR #425, opened 2020-05-11
Added handling for sparse precomputed distance matrices

Rocketknight1:
This is a PR for issue #424. I've confirmed that it works for my use case, but I'm not fully confident I haven't broken something else! Please double-check what I wrote before merging.

Also, I wasn't sure what datatypes to use for self._knn_indices and self._knn_dists, so I just used np.int and np.float. If you'd prefer int32/uint32 and float32, let me know!

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -22,7 +22,7 @@ except ImportError:
 
 import numpy as np
 import scipy.sparse
-from scipy.sparse import tril as sparse_tril, triu as sparse_triu, dia
+from scipy.sparse import tril as sparse_tril, triu as sparse_triu
 import scipy.sparse.csgraph
 import numba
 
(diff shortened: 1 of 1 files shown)


Rocketknight1:
Ah, sorry, my local line length is a bit longer than PEP8 wants. I'll fix that too when submitting any other changes later.

lmcinnes:
I think the real answer is that we currently prefer to just use black to solve all code formatting issues. If you get a recent version of black installed and run it on any files you touched that should resolve formatting issues, including the travis-ci failure (which is a check for black formatting).

umap/umap_.py
```diff
@@ -1688,8 +1689,43 @@ def fit(self, X, y=None):
...
+                raise ValueError("Sparse precomputed distance matrices should be symmetrical!")
+            self._knn_indices = np.zeros((X.shape[0], self.n_neighbors), dtype=np.int)
+            self._knn_dists = np.zeros(self._knn_indices.shape, dtype=np.float)
+            for row_id in range(X.shape[0]):
+                # Find KNNs row-by-row
+                row_data = X[row_id].data
+                row_indices = X[row_id].indices
+                # Remove distance from sample to itself (i.e. the matrix diagonal), if present
+                row_data = row_data[row_indices != row_id]
+                row_indices = row_indices[row_indices != row_id]
```
lmcinnes, line 1707: Usually the knn includes the point itself, and distances of zero are handled specially. I suspect this doesn't break anything, but I believe it is not necessary either (unless samples potentially have non-zero distances to themselves).

lmcinnes:
Thanks for this -- it looks good, and is definitely appreciated.

umap/umap_.py
Rocketknight1: I removed these and replaced them with a check that all the entries on the diagonal are either missing or zero.

Rocketknight1: Do you want me to explicitly insert zero values on the diagonal, to ensure consistency?

PR #427, opened 2020-05-15
Add text search widget to umap.plot.interactive to highlight matching points

radinplaid:
It's often useful to be able to filter/highlight ponts in a UMAP plot based on features of the original data. This functionality is implemented in this pull request.

By default, the behaviour is business as usual. But if `interactive_text_search` is set to True, a Bokeh widget is exposed just above the umap interactive plot and by default all columns in the bokeh data source can be searched. Points matching the search have their alpha (transparency) updated to interactive_text_search_alpha_contrast (default 0.95), and those that don't to (1-interactive_text_search_alpha_contrast) (0.05 by default)

The final parameter is `interactive_text_search_columns`, which lets users search a subset of columns in the data source. All columns in hover_data are available for searching as well as ['x','y','alpha','index','color'] and 'label' if there is one.

Highlighting is done via custom javascript as described here: https://docs.bokeh.org/en/latest/docs/user_guide/interaction/callbacks.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,8 +50,14 @@ configuration = {
         "tbb",
     ],
     "extras_require": {
-        "plot": ["pandas", "matplotlib", "datashader", "bokeh",
-                 "holoviews", "colorcet"],
+        "plot": [
+            "pandas",
+            "matplotlib",
+            "datashader",
+            "bokeh",
+            "holoviews",
+            "colorcet",
+        ],
         "performance": ["pynndescent >= 0.4"],
     },
     "ext_modules": [],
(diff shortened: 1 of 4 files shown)


jc-healy:
It looks like one build took >30 minutes to run instead of the usual 5 minutes then pynndescent failed to converge.
_AssertionError: 0.9782435129740519 not greater than or equal to 0.99 : NN-descent did not get 99% accuracy on nearest neighbors with callable metric._

This doesn't strike me as an error caused by Mark's code.

lmcinnes:
No, that's a nagging error that crops up sometimes; nothing to do with this PR. Thanks for this Mark! I'm sure a lot of people will find the very useful indeed.

PR #433, opened 2020-05-20
umap.plot.interactive text search now searches labels and hover_data …

radinplaid:
If you set interactive_text_search to False:
- It's pretty much the same as before (with the addition of 'alpha' to the data source)

If you set interactive_text_search to True but do **not** set interactive_text_search_columns:
- It searchs labels and/or hover_data
- If you don't provide labels or hover data, it warns you and doesn't display the search box.

If you set interactive_text_search to True and set interactive_text_search_columns:
- It searches the columns you asked it to search

### DIFF
--- a/umap/plot.py
+++ b/umap/plot.py
@@ -1298,7 +1298,7 @@ def interactive(
         Whether to include a text search widget above the interactive plot
 
     interactive_text_search_columns: list (optional, default None)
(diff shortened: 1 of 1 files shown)


lmcinnes:
Nice feature addition; thanks Mark!

PR #437, opened 2020-05-22
fix sparse_named_distances iteration

### LINKED ISSUE #436: Bug in sparse_named_distances iteration
See [here](https://github.com/lmcinnes/umap/blob/b2b6b3085e865273198fd3b0b41fef40d33d2e0e/umap/spectral.py#L107)
As we're iterating over a `dict` key-value pairs, `items` method should be called.

festeh:
Fixes #436


lmcinnes:
Thanks; good catch

PR #443, opened 2020-06-12
0.5dev densMAP

hhcho:
Adds support for densMAP based on the implementation in https://github.com/hhcho/densvis. README is updated accordingly with a reference, usage descriptions, and an MNIST example.


lmcinnes:
Hi @hhcho, thanks for this! Don't worry about the pep8 stuff, it is just being finicky. I'll try to work through this in the next week with feedback. I may be a little slow, but I am definitely very excited to have this integrated into UMAP!

lmcinnes:
This actually all seems fairly straightforward and easy to follow, so thanks for coding it up so cleanly. I have some minor concerns about performance and whether I want to actually break out an entirely separate code-path to retain optimal performance for the basic use-case of UMAP, but that's something I can benchmark and deal with if it comes up.

Thanks for taking the effort to add this -- I really feel like it is a powerful and useful addition to the package, and I hope to see it get lots of use in the future.

hhcho:
Great! I also wondered about having two separate code-paths but wanted to avoid duplicating the shared code. A regular run of UMAP should be nearly identical to before except for the additional arguments (with dummy values which are tiny) that get passed into the gradient update function; these are left untouched in the case of UMAP. I would guess that this overhead is negligible but if not splitting into separate functions makes sense to me.

Thanks again for inviting us to send a PR!

PR #445, opened 2020-06-15
Fix penguin images

ekerazha:


lmcinnes:
Thanks, it seems like they got moved around within the repository.

PR #470, opened 2020-07-23
added wrapper method UMAP._fit_embed_data 

timsainb:
This PR would add a wrapper method UMAP._fit_embed_data  around function simplicial_set_embedding in UMAP.fit that can be replaced in a child.


lmcinnes:
Errors are on me -- the dev branch is a bit messy right now.

PR #479, opened 2020-08-14
Error 404 on penguin dataset

christophfink:
The penguin data set is now available as an R file only, changed the URL to point to a previous commit’s file.


lmcinnes:
Thanks. I like the data set, but I have found the constant updates a little frustrating at times. I appreciate the help is keeping the docs working.

PR #486, opened 2020-08-27
Add Orion Search as a use case.

kstathou:
Hi @lmcinnes, thank you for developing UMAP! We are working on a [knowledge discovery tool](https://orion-search.org/) for scientific texts and UMAP is a major component of the web interface so I wanted to add it in the documentation as a use case. 

Let me know if I should make any changes!

Best,
Kostas


lmcinnes:
This looks like a great addition -- thanks for adding it!

PR #517, opened 2020-11-09
Fixed typos and minor rephrasing in the examples and user guide

Fietel:
Hey,
I fixed some typos/irregularities I found while reading the `Tutorial` and `Examples` sections.

The `benchmarking.rst` file seems to be outdated and replaced by `performance.rst` is that correct (I did not fully really check either yet)? 
The caption for the last figure in `docs/supervised.rst` also seems to be wrong. I think it should be "Test" instead of "Train", I didn't change it yet to remain consistent with the content of the figure itself, I can recompute the figure if you would like.

Thank you for this great project and documentation!


lmcinnes:
Thanks.

Issue #523, opened 2020-11-18
we should disconnect points beyond a certain distance


jc-healy:
So it seems that the UMAP assumption that our manifold is connected has a small problem when confronted with some interesting class of real world data.  It is particularly noticeable when dealing with bounded distances such as Jaccard, Hellinger, etc... 

In those distance measures you can (and do) get points what have no features in common with **any** other point and thus have a maximal distance of 1 to **every** other point.  We will calculate their k-nearest neighbours (a random selection of k equidistant points) and then apply the UMAP kernel to induce a similarity over those edges.  Because each of those points are equidistant (dist=1) they will all be called maximally similar.  Too many of these points will pull together random sets of points and distort the manifold you are trying to learn.

An easy solution to this would be to add a new parameter disconnection_distance to our UMAP model which prunes edges greater than that distance from our knn graph before applying the UMAP kernel.  This may trim some edges from some points and fully disconnect other points from our manifold entirely.   I'll add a variable severity warning depending on what proportion of your vertices are fully disconnected.  I propose having a dictionary of bounded metrics (DISCONNECTION_DISTANCES) which defaults our disconnection_distance to the maximal value for our bounded distance measures.  Otherwise it should default to np.inf (anything that a distance function declares as np.inf apart from something else should never be connected to it.  As we support new bounded metrics we should go through and add them to this dictionary.

This will result in disconnected vertices being embedded via the spectral intitalization (or random initialization) then slowly getting pushed away from all other points.

I'll add a umap.util.disconnected_vertices(model) function as well for easy post modeling access to which vertices were disconnected.  That should let people easily either filter them and re-run UMAP or just filter them out in the plotting via the subset_points functionality.

nrcjea001:
Hi @jc-healy. Apologies for my lack of understanding, but would you mind explaining how this parameter can be used (if at all) when using umap for outlier detection? When you say "points what have no features in common with any other point", could you also be referring to outliers? If so, can we say that this is an important parameter to tweak in the modelling?

jc-healy:
Hi there.  My short answer is that I *know* it is important in case of
bounded distances and we currently default it to 1 (max) for those
distances.  Such points are extreme outliers and typically have a
significant effect on an embedding.  If I missed one of the bounded
distances that we support then we should add a new default to the
dictionary of defaults.  If a user specifies a bounded custom distance then
they probably set the value to the maximal value of their distance
measure.  In other cases I'm less confident that it will be an important
[...]

PR #524, opened 2020-11-18
Dev0.5 disconnect distance (Fixes Issue #523)

jc-healy:
This PR should address the problems raised in issue: "we should disconnect points beyond a certain distance #523"


jc-healy:
Hold off on accepting this PR.  I realized that I still need to hammer out the right behaviour on the .transform() function for new points that are fully disconnected from our manifold.

PR #533, opened 2020-12-11
Fix for metric being mapped to metric name not valid for sklearn.

seinsinnes:
This is a work around for this problem I'm having:

  File "/usr/local/lib/python3.8/site-packages/fire/core.py", line 672, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "pipeline.py", line 41, in process_raw_text
    doc_to_vec_transform = umap.UMAP(n_components=2, metric='manhattan').fit(word_doc_matrix)
  File "/usr/local/lib/python3.8/site-packages/umap/umap_.py", line 1965, in fit
    self.embedding_ = simplicial_set_embedding(
  File "/usr/local/lib/python3.8/site-packages/umap/umap_.py", line 1033, in simplicial_set_embedding
    initialisation = spectral_layout(
  File "/usr/local/lib/python3.8/site-packages/umap/spectral.py", line 298, in spectral_layout
    return multi_component_layout(
  File "/usr/local/lib/python3.8/site-packages/umap/spectral.py", line 186, in multi_component_layout
    meta_embedding = component_layout(
  File "/usr/local/lib/python3.8/site-packages/umap/spectral.py", line 117, in component_layout
    distance_matrix = pairwise_distances(
  File "/usr/local/lib/python3.8/site-packages/sklearn/utils/validation.py", line 72, in inner_f
    return f(**kwargs)
  File "/usr/local/lib/python3.8/site-packages/sklearn/metrics/pairwise.py", line 1738, in pairwise_distances
    raise ValueError("Unknown metric %s. "
ValueError: Unknown metric taxicab. Valid metrics are ['euclidean', 'l2', 'l1', 'manhattan', 'cityblock', 'braycurtis', 'canberra', 'chebyshev', 'correlation', 'cosine', 'dice', 'hamming', 'jaccard', 'kulsinski', 'mahalanobis', 'matching', 'minkowski', 'rogerstanimoto', 'russellrao', 'seuclidean', 'sokalmichener', 'sokalsneath', 'sqeuclidean', 'yule', 'wminkowski', 'nan_euclidean', 'haversine'], or 'precomputed', or a callable

I ask for metric name manhattan but when it gets down to sklearn module it has changed into taxicab which sklearn doesn't recognise. This happens because the metric names are mapped to functions and then mapped back to names. This isn't a one-to-one mapping so it is possible for the metric name to be different from the one started with. Which means that it can go from a name which is valid for sklearn pairwise_distances to one which isn't.

Probably there is a better way to solve this but my current work around is to just create the inverse mapping with only those names which are valid for sklearn.


lmcinnes:
Looks like a perfectly acceptable workaround to me. Thanks for catching this -- it is certainly a quirky problem with the roundtrip from string to function to string. I'm just trying to track down why some tests are failing.

lmcinnes:
The test infrastructure got an overhaul that will likely clear some (or all of these problems). If you could resync your branch to the current master that may help a lot.

PR #538, opened 2020-12-23
Update aligned_umap_basic_usage.rst

EricThomson:
Changed `1797` to `digits.data.shape[0]` which seems more clear, and consistent with usage elsewhere in the notebook.


lmcinnes:
Good call, thanks!

PR #573, opened 2021-02-06
Fix for column names with whitespace in hoverdata

jgreene:
Bokeh requires escaping column names with { } as mentioned https://github.com/bokeh/bokeh/issues/4796 .


lmcinnes:
Thanks!

PR #582, opened 2021-02-11
add global structure preservation loss (MDS)

timsainb:
This PR adds a MDS loss encouraging correlation between distances in data space and embedding space, weighted by the parameter `global_correlation_loss_weight`. If global_correlation_loss_weight is 0 (default) do not add MDS loss.


lmcinnes:
The Azure coverage error is my fault, so we can ignore that for now. Let's get this merged :-)

PR #584, opened 2021-02-12
New Testing Pipelines with Py38 on CI

leriomaggio:
This PR includes new CI pipelines using Python 3.8 - so to keep aligned with CI in `pynndescent` :)


leriomaggio:
On this note, I have a suggestion: 

Considering that the pipeline here on Azure Pipeline is very thorough and goes on for many Py versions and SOs, I would suggest to drop other CI pipelines and just stick with Azure Pipelines - which also look to be the most efficient to complete.

Unless there is a reason to keep them which I am missing to understand, to me they look now useless duplications that require extra effort to keep updated :) 

my2¢ :)

lmcinnes:
The Azure pipelines was pretty new, and I was having a few issues getting it working as desired (including code coverage handling), so the others got left in place in the meantime. There is no particular reason not to eliminate them, although occasionally they surface errors not apparent in the azure pipelines (although that is always a source of puzzlement).

leriomaggio:
Great, thanks for the clarification. 

I am waiting to see how *this* whole pipeline will do and then we could take decisions about it.

Just saw that coverage failed. Googled quickly and found this: https://github.com/lemurheavy/coveralls-public/issues/1272

Maybe it could be useful to fix it? I am not sure I have permissions to do that on my end? 

As for other CIs:

At the moment, I can tell you that `Appveyor` is doing a pretty good job in testing on Win machines - and there is usually no surprise. Jobs on Wins on Azure have completed successfully. 
So, to me this looks like the first I would get rid of.

As for Travis, they recently changed pricing and I heard that some open source projects are migrating to either Azure Pipelines or Github actions (e.g. `scikit-learn`) 
Will wait and see, but if it turns out to be redundant, I would drop support to travis as well.

lmcinnes:
As a rule the azure coverage builds work for standard builds, but for builds of PRs they seem to fail; relating to tokens. I'll take a look at the issue you linked and see if that helps. Regardless I can just merge this at least.

Issue #590, opened 2021-02-16
[DEV] test on gradient metrics match fail on hellinger


leriomaggio:
Reporting this bug found in development for log purposes - I am sure @lmcinnes you're already aware.

I tried a couple of things very quickly - so my goal here is to give you _clues_ and possible _directions_ in case you haven't the time to look into that yet 🙌 

It seems that there is a `failure` in the `test_grad_metrics_match_metrics` when testing the default `hellinger` metric [L529](https://github.com/lmcinnes/umap/blob/05840ef919f2367d77e3ba929de86ea31056e691/umap/tests/test_umap_metrics.py#L529)

In particular, it seems that 
```python 
dist_matrix = dist.pairwise_special_metric(
        np.abs(spatial_data), np.abs(spatial_data)
    )
``` 
returns an empty distance matrix:

```python
>       assert_array_almost_equal(
            test_matrix,
            dist_matrix,
            err_msg="Distances don't match " "for metric hellinger",
        )
E       AssertionError: 
E       Arrays are not almost equal to 6 decimals
E       Distances don't match for metric hellinger
E       (shapes (10, 10), (0, 0) mismatch)
E        x: array([[0.      , 0.418969, 0.354731, 0.453094, 0.408693, 0.325863,
E               0.423645, 0.347026, 0.410423, 0.312178],
E              [0.418969, 0.      , 0.397624, 0.414187, 0.447527, 0.377329,...
E        y: array([], shape=(0, 0), dtype=float32)
```

The failure **does not** occur when `Y=None` in `dist.pairwise_special_metric` invocation, namely:
```python
dist_matrix = dist.pairwise_special_metric(np.abs(spatial_data))
```

**Symptoms and Clues**:

The failures resolves when either (A) `Y=None` **OR** `parallelization` is removed from the `numba.njit` decorator
([`distances.py:L1252`](https://github.com/lmcinnes/umap/blob/05840ef919f2367d77e3ba929de86ea31056e691/umap/distances.py#L1252))

This led me thinking that the failure may be caused by a possible *race condition* happening during parallelization which results in wrong calculations at `numba` level - _but I may be wrong!_
See [Numba Parallel - Explicit Parallel Loops](https://numba.pydata.org/numba-doc/dev/user/parallel.html#explicit-parallel-loops)

lmcinnes:
[...]

PR #591, opened 2021-02-17
First implementation to Fix Issue #590

leriomaggio:
Ok @lmcinnes so, following on your comment on the ugliness of the first workaround implementation (and I totally shared the feeling), here is a different attempt to fix the function and to also get rid of code duplications with a more compact implementation covering the two cases.

**Issue**: The issue was indeed related to the `result` array being allocated differently (as in different shapes) within the two alternative `if` branches - this upsets `numba` for some reason. Still not quite sure why.
Therefore, another quick _workaround_ implementation is provided, and tested as `pytest.fixture`
(please see [here](https://github.com/lmcinnes/umap/blob/abdfddb650d28dbae74be2202c370fde0ab67fee/umap/tests/test_chunked_parallel_spatial_metric.py#L7))

Specifically, a new _dedicated_ test module has been created to verify the **proposed** [compact implementation](https://github.com/lmcinnes/umap/blob/abdfddb650d28dbae74be2202c370fde0ab67fee/umap/distances.py#L1252) and the original workaround.

Of course, it goes without saying that both implementations allow all tests in `test_umap_metrics` to **PASS**.

Please see below the description of the commit:  

>This commit includes a new implementation for the `chunked_parallel_spatial_metric` in `distances` to fix issue #590 causing Python 3.8 to fail tests in the case when `X` and `Y` parameters are passed.
>
>In more details, this new implementation addresses the issue and also provides a more compact version to avoid code duplications on the two branches (i.e. `Y is|is not None`).
>
>In order to tests the two implementations, a special fixture/decorator is created to compare calculations with a quick-fix implementation of the previous original `chunked_parallel_special_metric` as implemented by Leland.

Feel free to send feedback on the compact implementation: personally I tried to replicate the same logic in your initial idea. 

However, I am not necessarily convinced that adding those `if clauses` for the `symmetrical` cases is indeed more performant than doing the iteration loop (in parallel) in all locations, anyway. 
It is indeed true that in the case of very big matrices, lots of computation can be saved - at the price of `if clauses` which are notoriously costly. 

I would love to hear your comments on this! Thank you :)


lmcinnes:
I like what you manage to do the clean things up from the first fix you posted to the issue! It looks a lot cleaner now. I'll have to do some benchmarking and see how performance compares because, as you note, the if statements now introduce some extra overhead.

lmcinnes:
So the good news is that it works; the bad news is that, at least in my quick benchmarking, it is very slow. I suspect the added if statements create a lot of overhead in what are tight inner loops, so the whole thing slows down a lot. I'll try and see if I can come up with an alternative formulation that still works, but runs as fast as the old version.

leriomaggio:
Great @lmcinnes thanks a lot for the feedback and the quick benchmarking - I was indeed too curious to see whether our initial guess about the `if-clause` were indeed correct. 

So I setup a quick benchmark tests (using `pytest-benchmark` - first time, lots of fun indeed!!) and run it on two alternative implementations (I simply tried the same getting rid of the if clauses). 

I could indeed try the "ugly" workaround too for comparison 👍 
I also pushed the benchmark report - thought it was the quickest to share it, we will remove it before merging.

Here is the link to the [benchmark report](https://github.com/lmcinnes/umap/blob/quickfix_chunked_parallel_special_metric/chunked_parallel_benchmark_report_valerio_macbook_CPU.txt)

**Note**: I've done a bit of editing to testing function names to allow for an easier-to-read report!
Quick legend: `test_with_IF_` refers to the implementation with the IF clause checks ([implementation](https://github.com/lmcinnes/umap/blob/d4b54b73ec57267fecad07cc513df94f4063f848/umap/tests/test_chunked_parallel_spatial_metric.py#L65)); 
`test_FULL_it_` refers to the alternative implementation with no control, and full iterations ([implementation](https://github.com/lmcinnes/umap/blob/d4b54b73ec57267fecad07cc513df94f4063f848/umap/tests/test_chunked_parallel_spatial_metric.py#L99)).
Suffices `_x_only` or `x_y` are self-explanatory and refers to the function called with either `X` and `X, Y` respectively.

Results are indeed interesting, now I'm intrigued how the workaround behaves. 
At a very first glance, it looks that we were right!

Looking forward to receiving your impressions too! 🙌 
Thanks a lot for testing on Py3.9 & Numba as well!

leriomaggio:
Ah btw, "RE: `pytest-benchmark`

Forgot to say: the great thing about that plugin was that I was able to specify a "group" name for tests so it automatically grouped runs for comparison. It seems very well done, indeed.

lmcinnes:
I wouldn't mind the "ugly workaround" for comparison, although your new alternative approach to remove the if looks pretty good.

The pytest-benchmark looks pretty useful! I wonder if we should have some benchmark reports produced for a few standard datasets so we can check for performance regressions. It would probably be a matter of setting that up as a separate run on azure pipelines. Certainly something for the future.

leriomaggio:
Brilliant! 

The "ugly workaround" is running - it _should_ take less time than the other, and I should be able to post results soon! 🙌 

RE: Performance Regression tests: ➕ 1 
I personally think that this is a wonderful idea! Should also allow to keep track of performance through the different `numba` versions in the backend.

leriomaggio:
Report pushed: the two implementations are indeed pretty close! (the ugly is still winning slightly)
In this benchmark I also changed the input data size, ranging from a `matrix [100x100]` up to `[1000x1000]`.

They ran on my laptop (with no GPU acceleration). Performance are not that bad, after all :)
Intrigued to test on a GPU - will try that later, if I have time.

lmcinnes:
The two options look pretty much comparable, so I would suggest you go with whichever you prefer.

leriomaggio:
Brilliant, thanks! So I already updated `distances.py` with the new implementation in my last commit. 
The last thing I would like to discuss with you is what to do with the test and the reports in the repository, before merging into `main`.

I shall assume I can get rid of performance reports, and that's fair.
What about performance testings? 

My suggestion right now would be to keep them in the test suite, but skip them - similarly to what we do now when skipping tests for unmet dependencies (e.g. `Tensorflow`).

So, if no `BENCHMARK` variable is set in `os.environ`, we are going to skip those tests - so that we can enable that variable in the future if we aim at including performance regression testing. 

Does this make any sense?

Thanks, Valerio

lmcinnes:
Sounds good to me!

leriomaggio:
> Sounds good to me!

Brilliant, thanks! I've just pushed those changes, and a bit of refactoring - polishing the tests a bit. 
It should be ready for MERGE :)

PR #600, opened 2021-02-23
Fixing flaky test

sleepy-owl:
The test `test_aligned_update` fails non-deterministically sometimes. This PR fixes this issue.

To find a solution, I collected samples of the accuracy value from several test executions and computed the tail distribution. I computed the extreme percentiles to check how high can the values be. 

```
0.99::0.63
0.999::0.56
0.9999::0.45
```

I selected the 99.99th perc for the fix. I think setting the bound using the statistical evaluation might be a good way to ensure the test is not flaky.

Do you guys think this makes sense? Please let me know if this looks good or if you have any other suggestions. Also, here I assume there are no bugs in the code under test.


lmcinnes:
Thanks -- the non-deterministic tests are a little annoying at times. I appreciate the effort you went to to ensure this won't trip accidentally.

Issue #609, opened 2021-03-10
(semi-)supervised + densmap


rxjx:
Is there a theoretical reason why densmap is not allowed for supervised or sem-supervised mapping?

lmcinnes:
I don't believe so, no. It may, however, be rather quirky given conflicting requirements about local density.

rxjx:
Yeah, I can see how the more complicated objective function would become difficult to balance. It would be useful though to see the local density vs supervision trade-off. Might help understand how shattering of the manifold helps with classification. Probably a fertile area for lots of papers :-)

lmcinnes:
I'm pretty sure you can just comment out these lines:
https://github.com/lmcinnes/umap/blob/b1223505ca56ae104feb35e4196227277d1e8058/umap/umap_.py#L2467-L2470

and it will run densmap in supervised mode and give you the quirky results, so if you are interested in exploring that it may well be worth taking a look. One could also try to work out how to adapt the local density estimates to be better in the supervised or semi-supervised case (estimate the high dimensional local density only for points in the same class? for points not in a different class for the semi-supervised case?) I don't have time to play with that now, but if you find it does work well, or can come up with workable adjustments I would certainly welcome a PR.

rxjx:
Thanks, I'll try it out when I can. Free time is the problem :-( but I hope to steal some to look at this.

rxjx:
Yeah, uncommenting those lines worked just fine. Will get a PR to you soon.

PR #616, opened 2021-03-12
Allow densmap for supervised embedding

rxjx:
See https://github.com/lmcinnes/umap/issues/609


rxjx:
Ahh, the test needs to change. Sorry!

PR #617, opened 2021-03-12
Densmap supervised (issue 609)

rxjx:


lmcinnes:
Could you run it supervised on Fashion-MNIST and post the resulting plot? I just want to be sure it "mostly works" in reasonable cases.

rxjx:
Yeah, I think I can do it over the weekend.

rxjx:
image

lmcinnes:
Looks good! I was worried there might be enough weird interactions to make it go screwy, but it seems good enough for now. Thanks for taking the time and trouble to look into this. The test is also greatly appreciated.

rxjx:
No problem at all. Thank you for UMAP and all the work you've put into it.

PR #620, opened 2021-03-15
Implement masking to control how embedded points are updated

matthieuheitz:


umap/umap_.py
```diff
@@ -2671,7 +2727,7 @@ def fit_transform(self, X, y=None):
         r_emb: array, shape (n_samples)
             Local radii of data points in the embedding (log-transformed).
         """
-        self.fit(X, y)
+        self.fit(X, y, pin_mask)
```
lmcinnes, line 2730: Probably best to make this a keyword arg in case we get more fit parameters later (such as a ``init`` fit param).

```suggestion
        self.fit(X, y, pink_mask=pin_mask)
```

lmcinnes (commented):
Looks good so far. I would like to consider different names than ``pin_mask``. It is a small thing, and naming is hard (I don't have a particularly good alternative suggestion), but good names help users a lot. My best suggestion at the moment is ``inertia`` or ``sample_inertia`` (similar to ``sample_weight`` used in other sklearn models). Of course we would need then then pass through ``1.0 - inertia`` but that seems tractable. I would welcome your thoughts however.

lmcinnes:
Note that the remaining test failures are due to how coveralls plays with azure, so you can safely ignore them.

Another note: I may start a new 0.6dev branch and target this at that so we can merge it in sooner and work on it more easily without messing with the main branch. Let me know if you think that would be a good idea.

jondo:
I have now tested this, and I found out that I additionally needed [to block this rescaling](https://github.com/jondo/umap/commit/c750482455f182ddb2e265fa829f9cc0263c82c5) such that my points (which have coordinates slightly outside [0, 10]) stay fixed.

jondo:
I also think that warning instead of rescaling is the way to go. Perhaps "one of the coordinate _ranges_ is outside [8, 12] (i.e. more than 20 % off)" is a semi-reasonable condition? This assumes that the layout optimization is independent of absolute embedding location.

kruus:
A colleague working on an interactive visualization tool had an interesting scenario.  A user could (say) inspect and then drag certain points to left/right and "pin" the x-axis of the dragged points while leaving other axes of the dragged points adaptable.

Supporting also a 2-D nsamples x dim `pin_mask` seems to describe the capability he was looking for.
In this case, we want the behavior of gradient weights `other_mask` and `current_mask` to change from scalars into vector gradient multipliers.

Issue #606, opened 2021-03-04
Possibility to fix points in the low embedding


matthieuheitz:
Hi,
I'd like to know if it would be possible to add a prior on the low embedding, meaning that I want certain points to be at certain coordinates in the low embedding, and I would like UMAP to optimize all other points.
Is that something that is doable with the actual code, or would it need some changes?
How difficult would it be to add this feature?
Thank you very much.

lmcinnes:
The code, as it stands, would need some changes to allow this to happen. This is actually already a requested feature (see #432) -- although that request may differ in specific details compared to what you need. It was on my development plan, but I have not gotten to it yet. It is not really that difficult to add, so if you wanted to have a look at doing it yourself I can provide guidance as to what changes would need to be made.

matthieuheitz:
Yes, that would be fantastic!
I could definitely add this with some guidance.
My initial thought was also to have a mask that says which points are fixed.
Do you have a preferred way for communication? Discussing all details here might be too much.

lmcinnes:
You can email me at "leland DOT mcinnes AT gmail DOT com" and I can give you a general outline and we can discuss from there.

jondo:
@matthieuheitz, thank you very much for tackling this! How is it going so far?
Actually, I would be interested to have the detailed discussion here.

matthieuheitz:
Yes, it's not complete yet, and it's still breaking some CI checks, so I'll work on that this week.
But yes, you can already look at my changes on the PR.

jondo:
In the meantime I have tested @dribnet's brute force implementation from https://github.com/lmcinnes/umap/issues/432#issuecomment-633145366 ([see my code](https://github.com/jondo/umap/commit/3f3fe1ea31c0d1dd3d8a1c4ca966d5dd2e2f0e2d)).

I have now first computed an embedding _without pinning_, then I saved some resulting data point coordinates from there. Of course, computing another embedding (with same data and random seed) by pinning these data points leads to an overall different embedding (except for the pinned points).

I now wonder whether there could be a _stable_ implementation of pinning, s.t. pinning data points to their current embedding locations would not change the embedding. Maybe one could internally first compute the unpinned embedding, then do an additional "pinning adjustment" phase?

jondo:
[...]

PR #634, opened 2021-03-30
Enable n_epochs=0 to get the initial embedding

jondo:
This extends the range of possible `n_epochs` values below the current limit 11.

It is e.g. useful to examine the computed spectral embedding. Also, the resulting embedding can then be [masked to fix points](https://github.com/lmcinnes/umap/issues/606).

Note that for the new value range, in `simplicial_set_embedding` the graph data is only thinned with the `n_epochs` default value, otherwise spectral embedding would give `UserWarning: Graph is not fully connected, spectral embedding may not work as expected` (or even loop forever for `n_epochs=0`).

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -1044,18 +1044,24 @@ def simplicial_set_embedding(
     graph.sum_duplicates()
     n_vertices = graph.shape[1]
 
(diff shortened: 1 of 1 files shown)


jondo:
Thank you for merging this!

PR #639, opened 2021-04-01
Support predicting new points with precomputed distances

Rocketknight1:
This might be too niche, in which case feel free to reject it!

This is a quick PR that enables predictions for new points when the metric is precomputed. We assume in this scenario that the provided matrix is in fact distances from the new points to the original points in the training set. This means that the shape of the provided matrix should be (num_new_points, num_training_points).

The matrix can be sparse - in this case, we assume that the nonzero elements of the matrix include the n_neighbors nearest neighbors, or at least the approximate nearest neighbors. We also print a big honking warning so people know what's expected when they try to do this.

### DIFF
--- a/umap/tests/test_umap_on_iris.py
+++ b/umap/tests/test_umap_on_iris.py
@@ -4,7 +4,7 @@ import numpy as np
 from sklearn.cluster import KMeans
 from sklearn.metrics import adjusted_rand_score
 from sklearn.neighbors import KDTree
-from scipy.spatial.distance import cdist, pdist
+from scipy.spatial.distance import cdist, pdist, squareform
 
 try:
     # works for sklearn>=0.22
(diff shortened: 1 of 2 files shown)


lmcinnes:
Definitely not too niche -- other people have asked for this but I hadn't gotten around to implementing it yet.

Any chance you could add a test that at least exercises the code? Potentially just copy one of the existing tests of the transform functionality and adapt it to run with precomputed distances.

Rocketknight1:
Sure! I'll try to get that in by tomorrow.

Rocketknight1:
It's in! I'm getting an odd error about Coveralls tokens in your checks, but all the tests pass for me.

lmcinnes:
Thanks! The failing tests are just configuration issues with azure and coveralls -- they will disappear once this gets merged in. Thanks getting this done, and adding the tests!

Issue #613, opened 2021-03-12
Addition of `alpha` argument in plot.point function


itsnamgyu:
I'd like to add transparency to the dots in the `plot.point` function. Would you be open to a PR with this feature implementation. It will be a simple change.

## Background

I'm trying to visualize the UMAP of a very large dataset and there are some very dense clusters in the embedding space that I would like to check visually. I could raise the resolution, but it's already very big. I've found that semi-transparent dots(?) is a simple yet useful solution.

lmcinnes:
I would certainly welcome a PR and we can discuss any finer details there. My only concerns are that I would like to try to keep the plot API simple with as few parameters as possible. There is also the question of how an alpha parameter would interact with the datashader plotting used for large/dense dataset. None the less, I see you point that this could very well be useful. Hopefully we can come up with something reasonable in the PR.

Issue #491, opened 2020-09-11
umap.plot.interactive broken in 0.4.6?


rmitsch:
I'm currently trying to use `umap.plot.interactive` with `umap-learn==0.4.6` by following https://umap-learn.readthedocs.io/en/latest/plotting.html. I want to dump the resulting vis. either to disk or display it in the browser directly, without using a notebook. 

However, neither `umap.plot.output_notebook()` nor similar functions exist in the current UMAP version. Furthermore, the output of `umap.plot.interactive` is of type `holoviews.core.spaces.DynamicMap` - `umap.plot.show()` expects an instance of type bpl.Figure. Even calling `show_interactive()` directly is not successful and results in `RuntimeError: no display hook installed for notebook type None`.

What is the correct way to plot with `umap.plot.interactive`?

lmcinnes:
I think this means that there was enough data that a different code-path (using holoviews and datashader) had to be taken. I will have to look into how to display those outside of a notebook; obviously the Bokeh calls are not cutting it. I would suggest you try the holoviews documentation or issue tracker for suggestions as to how to get it working in the meantime.

rmitsch:
Will do. Thanks for looking into it!

yeus:
still not working in umap 0.5.0

PR #645, opened 2021-04-11
Add alpha parameter for points and interactive methods in umap.plot

gclen:
This is an initial PR for https://github.com/lmcinnes/umap/issues/613. I added an optional alpha parameter for the points and interactive methods. I intentionally ignored adding it to the interactive plots for large datasets where holoviews is used. This is due to the fact holoviews/datashader has a min_alpha parameter which is used if alpha < min_alpha. If we want to support this we would also need to set min_alpha appropriately but that might be more trouble than it's worth IMO. I also fixed a bug where the holoviews plots wouldn't display in a Jupyter notebook and were throwing deprecation errors.

### DIFF
--- a/umap/plot.py
+++ b/umap/plot.py
@@ -259,6 +259,7 @@ def _datashade_points(
     width=800,
     height=800,
     show_legend=True,
+    alpha=255,
 ):
 
     """Use datashader to plot points"""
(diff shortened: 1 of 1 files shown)


gclen:
This should also address #491. I can pull that change out into a separate PR if you would like

lmcinnes:
Looks good. Thanks!

PR #661, opened 2021-05-01
Add plotting method for average NN distance

gclen:


lmcinnes:
A useful addition; thanks!

PR #667, opened 2021-05-10
Optionally enforce exact distances

jondo:
Since the threshold of 4096 for deciding the distance algorithm is a bit arbitrary and depends on machine speed, this pull request introduces another argument `force_exact_distances`as an opposite to the existing `force_approximation_algorithm`.


jondo:
@lmcinnes , this sounds plausible, but I won't do it now. Feel free to close this request.

PR #670, opened 2021-05-17
Bug fixing in Clustering documentation

nicolebussola:
change fetch_openml to load MNIST instead of Fashion-MNIST as the rest of the example refers to the MNIST data


lmcinnes:
Thanks!

PR #687, opened 2021-06-05
Add matplotlib animated GIF demo to aligned UMAP doc

chrismbryant:
Animation feels like a natural way to view aligned UMAP results, so I thought it made sense to include a demo for creating a GIF in the aligned UMAP basic usage docs. Please let me know if this is not the appropriate place to include such a demo.

Also, I added the GIF demo to the first plotting section, but maybe it would make more sense to include it at the very end instead as an alternative to the 3D plot?


chrismbryant:
Closing this, because PR #688 is better.

PR #688, opened 2021-06-06
Add matplotlib animated GIF demo to aligned UMAP doc

chrismbryant:
Related to PR #687.
> Animation feels like a natural way to view aligned UMAP results, so I thought it made sense to include a demo for creating a GIF in the aligned UMAP basic usage docs. Please let me know if this is not the appropriate place to include such a demo.

I added a GIF demo to the very end of the doc, since I think it works better both organizationally and visually. Using the interpolation from the 3D plot allows for very smooth motion in the animation. (I also updated both the related prior plots so that they all have a consistent random state)


lmcinnes:
Thanks! I had been using animations myself, but hadn't found a good clean way to insert them in the docs. This looks like a very effective approach.

PR #690, opened 2021-06-08
Fixed typo in warning

jamestwebber:
Since I've been seeing this warning a lot recently, thought I'd fix the typo in the message...`method` was repeated


jamestwebber:
I know! But there are tons of other examples in the code so I figured ¯\_(ツ)_/¯

lmcinnes:
Don't worry about the PEP warnings; I'm not so fussy about line lengths, but it is useful to catch some other things occasionally. Thanks for this. Small things like this do make a big difference.

PR #696, opened 2021-06-16
Get intermediate results at different epochs

matthieuheitz:
This is a proposition to be able to get intermediate results of the optimization, at different epochs.
You just provide a list of int to `n_epochs`, and the result will be available after having called fit(), in the UMAP object as the member `embedding_list`.

Usage:
```python
reducer = umap.UMAP(n_neighbors=n_neighbors, min_dist=min_dist, n_epochs=[100,200,300])
final = reducer.fit_transform(X)
embeddings = reducer.embedding_list  # List of size 3, containing embeddings at epoch=100, 200 and 300
# The last elements of `embeddings` is the same as `final`
```

For now, I've only implemented this for `optimize_layout_euclidean`. I'm waiting for your feedback on this one, and when it's good, I'll extend it to the other layout functions.


umap/umap_.py
```diff
@@ -1707,10 +1717,18 @@ def _validate_parameters(self):
                 raise ValueError("n_components must be an int")
         if self.n_components < 1:
             raise ValueError("n_components must be greater than 0")
+        self.n_epochs_list = None
+        if isinstance(self.n_epochs, list):
+            if not all(isinstance(n, int) and n >= 0 for n in self.n_epochs):
+                raise ValueError("n_epochs must be a nonnegative integer "
+                                 "or a list of nonnegative integers")
+            self.n_epochs_list = self.n_epochs
+            self.n_epochs = max(self.n_epochs_list)
```
lmcinnes, line 1726: Scikit-learn has a requirement that attributes that are params at creation time should not be changed by fitting/etc. That means you shouldn't change the value of ``n_epochs``. The usual practice is to have a separate value, possibly private like ``self._n_epochs`` that you use. You will see other params for UMAP using this approach.

umap/umap_.py
```diff
@@ -2579,12 +2597,22 @@ def fit(self, X, y=None):
...
                 random_state,  # JH why raw data?
             )
+
+            if self.n_epochs_list is not None:
+                if "embedding_list" not in aux_data:
+                    raise KeyError("No list of embedding were found in 'aux_data'. "
+                                   "It is likely the layout optimization function "
+                                   "doesn't support the list of int for 'n_epochs'.")
+                else:
+                    self.embedding_list = [e[inverse] for e in aux_data["embedding_list"]]
```
lmcinnes, line 2614: A Scikit-learn convention is that attributes set/created by a ``fit`` call should end in an underscore, so this should be ``self.embedding_list_``. Possibly you also want to use ``self.n_epochs_list_`` as well.

lmcinnes:
This looks really promising and I'm sure it will be useful t many people, so thanks for submitting it. I left some comments mostly just around scikit-learn conventions which is is useful to try to hew to wherever possible as it results in least surprise for users.

umap/umap_.py
matthieuheitz: Hey! Sorry for the delay.
I wanted to modify `self.n_epochs` so that I don't have to change the code in other places where it expects it to be an int, and so that future developers don't have to worry whether it's a list or not.
I could do just `self.n_epochs_max = max(self.n_epochs_list)`, but I wouldn't really use anywhere else.
Is it okay to just remove that line (`self.n_epochs = max(self.n_epochs_list)`) and let `self.n_epochs` being either an int, None or a list?

umap/umap_.py
matthieuheitz: Okay. But for `self.n_epochs_list`, it hasn't been created by fit(), it's just in the init, no?

lmcinnes:
Sorry for being so slow to circle back to this. It looks good. I'll try to have a final review soon, and then hopefully just merge it. Thanks again for the contribution.

PR #734, opened 2021-07-22
Reproducibility for Aligned UMAP

mattsmart:
Propagate a rng state through UMAP alignment using self.transform_seed as the seed.

Resolves Issue: aligned UMAP embeddings are not reproducible using random_state or transform_seed attributes.


lmcinnes:
Looks good. Thanks for this!

mattsmart:
No problem, thanks for checking!

PR #21, opened 2017-11-16
Basic logging added

sauln:
Here is a first pass on #18.  I have tested this in a notebook and a script.

I went with scikit-learn style logging ([example can be seen here](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/cluster/k_means_.py)) instead of the gensim style as mentioned previously. If you'd rather have the python module be used, we can convert it. There might be issues getting it to work within the `numba.njit` decorated functions though.

**Example usage:**
```
import umap
from sklearn.datasets import load_digits

digits = load_digits()
embedding = umap.UMAP(verbose=True).fit_transform(digits.data)
```

**Output:**
```
UMAP(n_neighbors=15, n_components=2, metric='euclidean', gamma=1.0, n_edge_samples=None, alpha=1.0, init='spectral', spread=1.0, min_dist=0.1, a=None, b=None, random_state=None, metric_kwds={}, verbose=True)
Construct fuzzy simplicial set
	nn descent iteration  0  /  10
	nn descent iteration  1  /  10
	nn descent iteration  2  /  10
	nn descent iteration  3  /  10
Construct embedding
	embedding iteration  0  /  11000000
	embedding iteration  1100000  /  11000000
	embedding iteration  2200000  /  11000000
	embedding iteration  3300000  /  11000000
	embedding iteration  4400000  /  11000000
	embedding iteration  5500000  /  11000000
	embedding iteration  6600000  /  11000000
	embedding iteration  7700000  /  11000000
	embedding iteration  8800000  /  11000000
	embedding iteration  9900000  /  11000000
```

PR #739, opened 2021-07-27
Fixes #18 and improves #21

### LINKED ISSUE #18: Add progress log
It would be nice to report some intermediate calculation progress as log info messages. For large data the process can be relatively long and progress info will help.

parashardhapola:
Hi @lmcinnes,

This PR improves the aesthetics of the progress log by using `tqdm` progress bar. The progress bar is only shown when `verbose` is set to `True` in the `UMAP` class.

Here is a screenshot of running the digits example data in **Jupyter lab environment**
umap_tqdm_example_jupyter

Here is a screenshot of running the digits example data in **ipython shell**
umap_tqdm_example_shell

The changes have been made to the following functions in `layouts.py` and I have added an extra parameter `tqdm_kwds` to each of them:
- `optimize_layout_euclidean`
- `optimize_layout_generic`
- `optimize_layout_inverse`
- `optimize_layout_aligned_euclidean`

`tqdm_kwds` is dictionary that contains keyword arguments for `tqdm`. Users can use this to override the look and
feel of the progress bar as per their need. Using `tqdm_kwds` it is also possible to disable the progress bar even when `verbose` is set to `True` (could be useful when logging UMAP output).

If you think this is useful then I can continue to work on this and expose the `tqdm_kwds` parameter all the way upto the UMAP class.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ scikit-learn>=0.22
 numba>=0.51.2
 pynndescent>=0.5
 tbb>=2019.0
+tqdm
--- a/umap/layouts.py
+++ b/umap/layouts.py
@@ -865,9 +865,7 @@ def _optimize_layout_aligned_euclidean_single_epoch(
 
                     for offset in range(-window_size, window_size):
                         neighbor_m = m + offset
-                        if (
-                                n_embeddings > neighbor_m >= 0 != offset
-                        ):
+                        if n_embeddings > neighbor_m >= 0 != offset:
                             identified_index = relations[m, offset + window_size, j]
                             if identified_index >= 0:
                                 grad_d -= clip(
@@ -887,9 +885,7 @@ def _optimize_layout_aligned_euclidean_single_epoch(
 
                         for offset in range(-window_size, window_size):
                             neighbor_m = m + offset
-                            if (
-                                    n_embeddings > neighbor_m >= 0 != offset
-                            ):
+                            if n_embeddings > neighbor_m >= 0 != offset:
                                 identified_index = relations[m, offset + window_size, k]
                                 if identified_index >= 0:
                                     grad_d -= clip(
--- a/umap/layouts.py
+++ b/umap/layouts.py
@@ -141,6 +141,8 @@ def _optimize_layout_euclidean_single_epoch(
                 grad_d = clip(grad_coeff * (current[d] - other[d]))
 
                 if densmap_flag:
+                    # FIXME: grad_cor_coeff might be referenced before assignment
+
                     grad_d += clip(2 * grad_cor_coeff * (current[d] - other[d]))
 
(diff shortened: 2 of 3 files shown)


parashardhapola:
So tests are failing when using `optimize_layout_generic` and `optimize_layout_inverse` functions because these are compiled by Numba. There are two options here:
- I either try to create something an equivalent of `_optimize_layout_euclidean_single_epoch` for both of above functions.
- Or we remove `tqdm` from above two functions and retract the changes introduced

parashardhapola:
In the second commit  I have added  `_optimize_layout_generic_single_epoch` and `_optimize_layout_inverse_single_epoch`. This make sure the the 'epoch loop statement' is outside numba's context.

This was a very ugly hack so might need an extra careful review. I'll be happy to followup.

lmcinnes:
This looks really good! Thanks for the work. I'll try to get some time in the next few days to go through it all and provide any feedback.

parashardhapola:
Great! I will continue to work on this through this week to expose the `tqdm_kwds` parameter to user level API

lmcinnes:
Let me know when you feel it is ready to merge. Also it would be nice if you could post some basic performance comparison benchmarks with and without the tqdm. I don't imagine the PR has any performance impact, but it would be nice to see some numbers.

parashardhapola:
Hi @lmcinnes,

This feels ready for review and merging.
Here is a [Github gist](https://gist.github.com/parashardhapola/0c698226437237718abd8da882cd1fda) for usage and runtime benchmark

Unfortunately,  Github does not render the notebooks with ipywidgets very nicely.
Here is an [HTML version](https://gist.github.com/parashardhapola/10bf08e15dc16373d9e5acc9139efa3a) of the notebook that can be viewed locallly.

lmcinnes:
Looks good to me; just a couple quick questions before I merge it. Are there any tests you think could be added for this? In practice I don't see too much -- most of the code will be exercised well by existing tests, but you may have a better sense. Also, I was hoping you could compare this patched version with the baseline current PyPI version for timing (just run a single benchmark script in different environments with different versions installed). Mostly I see this as not an issue, but if you have the time I would appreciate it (if you don't; don't worry, I'll take care of it myself later).

parashardhapola:
Hi @lmcinnes,

I will run some runtime benchmarks and come back to you shortly.
`tqdm` claims to add an overhead of 60 nanoseconds per iteration. So in most cases the expected overhead of the progress bar will be less than few tens of microseconds.

As for the unit tests. I too cant see how we can write tests specifically for the `tqdm` progress bar. All the altered code is already be well covered by existing tests.

lmcinnes:
Thanks; I'll merge this now, but I would still be interested in the benchmarks just for absolute peace of mind.

Thanks so much for this -- it will be appreciated by many, I suspect, and was well written so nice and easy to review.

PR #758, opened 2021-09-09
Fixed bug in umap.plot

pujaltes:
Calling umap.plot.connectivity() with a umap object with disconnected vertices caused python to crash. I found that the origin of the error was the _get_extent() function that used np.min and np.max to find the bounds for a plot. When there are disconnected vertices np.min(points)=np.max(points)=np.nan.  

This was fixed by calling np.nanmin() and np.nanmax() instead which ignore nan values.


lmcinnes:
Good catch, thanks!

Issue #786, opened 2021-10-14
Possibility of using precomputed knn


pujaltes:
Hi,
I'd like to know if you'd be open to adding the option of passing a precomputed_knn = (knn_indices, knn_dists, knn_search_index). In many applications one may want to test various values of n_neighbors. However, particularly for large datasets and high values of n_neighbors, this can become computationally expensive. I propose that we allow the user to pass a precomputed_knn when constructing a UMAP object. This precomputed_knn could be obtained by running the nearest_neighbors() function, or from a previously fitted UMAP object. 

Usage:
```
neighbors=[20, 50, 80, 100, 125, 150, 200]
X=np.random.randn(200000, 100)
embeddings = np.zeros((len(neighbors), X.shape[0], 2))

knn = umap.umap_.nearest_neighbors(X, n_neighbors=200, "euclidean", None, None, None)


for count, k in enumerate(neighbors):
    embeddings[count] = umap.UMAP(n_neighbors=k, precomputed_knn=knn).fit_transform(X)
```

In some cases we have also computed the knn before running UMAP to get a better idea of the nature of the data. Having this option could save users time when running UMAP with different parameters or when they already computed the knn previously. 

Thanks for taking a look and let me know what you think.

pujaltes:
Ok, sounds good. I'll give it a shot!

PR #788, opened 2021-10-17
precomputed_knn Implementation

pujaltes:
Hi,

As discussed in #786  I have added the option for the user to pass a precomputed_knn when constructing a UMAP object in order to save time. The precomputed_knn must be a tuple containing (knn_indices, knn_dists, knn_search_index). The first part of thisPR has to do with umap_.py:

1. Added docstring and basic implementation. I chose to create knn_dists, knn_indices and knn_search_index attributes to keep the precomputed_knn seperate from the ones that umap uses internally (with '_' before them).
2. Included several checks to validate user inputed parameters. 
3. In order to ensure reproducibility when using precomputed_knn and comparing with regular UMAP. Note that this is only ensured when the k of the precomputed_knn is equal to the n_neighbors value. This doesn't add any issues as the NNDescent() function has it's own input validation to ensure that the random_state input is proper. However, I'm not sure if it's worth it.

Secondly I added a notebook testing and showing some of the functionality of the precomputed_knn. The first part of the notebook deals with reproducibility and input validation, and I included it just to show that it works. I'll remove this section before merging. The second part of the notebook actually includes a use case example, and I think it would be useful to include it with the rest of the notebooks. 

I planned to continue working on some more examples, but, I wanted to see what you thought before proceeding.


jc-healy:
I'm excited to see this folded into the code base as well.  It will make a very nice addition to the library.  Well done!

pujaltes:
Hi, thanks for your comments. I'm glad you both liked it.

I hadn't considered how the change would affect previously reproducible UMAP results.  The change isn't entirely required to reproduce results. Precomputed k-nn's will produce the same result as long as the same seed and n_neighbors is used in the n_neighbors function. The only scenario where the change helped was if you wanted to reproduce regular UMAP with a precomputed_knn UMAP, or vice versa. 

However, after thinking about it I don't see this being all that useful. The precomputed_knn has to have the same number of neighbors as the normal UMAP. The only case where I can think of this being applied, is if you want to analyze the k-nn distribution before running UMAP with n_neighbors=k. I agree that we can look into updating it later, but, it doesn't seem to be worth it for now.

I set up a different branch where I removed the changes to the random_state variable. In it, I also polished the [notebook](https://github.com/pujaltes/umap/blob/knn_dist2/notebooks/precomputed_knn%20use%20cases.ipynb) in that branch to be a bit more didactic and I went into more detail about reproducing results with precomputed_knn. It turns out that even without changing when the random_state variable is declared, we can still reproduce normal UMAP results with a precomputed_knn. 

I'm very interested in seeing what you think. If you like the new notebook and modifications to the implementation, i'll merge the[ knn_dist2 branch](https://github.com/pujaltes/umap/tree/knn_dist2) to the current one in the PR.

lmcinnes:
The new notebook looks great. If you want to merge that branch into the PR I will be happy to merge it now. 

If you want to go a step further I would recommend adding the notebook explicitly to the documentation. You can do that by using ``nbconvert`` to generate an ``rst`` file, and putting that in the ``doc`` folder, then adding the name of the file (less the extension) to the ``doc/index.rst`` file (in the tutorials section would be best). Having it explicitly in the docs will make it much more likely that users will find it.

I'm happy with whatever you want to do from here -- just let me know. And great work, this is definitely appreciated.

pujaltes:
Thanks! 

I liked the idea of including the notebook in the doc and ended up moving it there. Let me know if there's anything else you need me to do.

Issue #801, opened 2021-11-13
AlignedUMAP doesn't use labels in .fit and .update


gclen:
The .fit and .update methods of AlignedUMAP accept y as a parameter but it's not actually used. I'll submit a PR to fix this.

lmcinnes:
Done. Thank you!

PR #802, opened 2021-11-13
Use labels in AlignedUMAP if provided

gclen:
Fix issue #801


lmcinnes:
Looks like all the test failures are due to some tensorflow/keras changes that have broken parametricUMAP; I'll just merge this since it looks like a very simple change.

Issue #805, opened 2021-11-16
Invalid escape sequence spam...


brilee:
I'm seeing lots of warnings of the following kind when I import the umap library:

umap/distances.py:102: DeprecationWarning: invalid escape sequence \s
umap/distances.py:117: DeprecationWarning: invalid escape sequence \m
umap/distances.py:131: DeprecationWarning: invalid escape sequence \m

I believe this can be fixed with an `r` prefix on all the triple-quoted docstrings that have inline LaTeX containing backslashes.

lmcinnes:
Yes, thanks. If you want to make a PR I would gladly merge it; otherwise I will hopefully get to this at some point soon,

brilee:
for future reference, to reproduce the escape sequences, run the following under python<=3.8: `rm -r umap/__pycache__/ && python -Wd -c "import umap"`

PR #807, opened 2021-11-20
Properly escape invalid escape sequences.

brilee:
Fixes https://github.com/lmcinnes/umap/issues/805


lmcinnes:
Thanks for this. The errors seem to relate to some updates in tensorflow interacting badly with parametric UMAP which I will have to track down at a later date.

PR #822, opened 2022-01-04
Add Mutual-NN Description

adalmia96:
Adding section for using a mutual-NN rather than a k-NN with UMAP based on this paper: https://arxiv.org/abs/2108.05525


lmcinnes:
Thanks!

dribnet:
Is this documentation out of date? I'd like to do some experiments with mutual-NN and would potentially be interested in helping make this a better supported option with sensible defaults.

lmcinnes:
@dribnet I think there is definitely more that could be said about this, and relative support for it. Right now a lot of this is essentially handled with the ``set_op_mix_ratio`` parameter which is not really well documented at all (or even that well named for anyone not familiar with the internals). I'd certainly welcome any suggestions or additions you have.

PR #829, opened 2022-01-17
Add missing links for graph layout methods

jlmelville:
Some links were missing in the How UMAP Works doc around some dimensionality reductions methods. I have added links to wikipedia articles.


lmcinnes:
Thanks!

PR #861, opened 2022-05-06
Fix AlignedUMAP.update

hndgzkn:
- Fixes two bugs in `aligned_umap.update`:
    -  if a list of values is specified in `aligned_umap.fit` for parameters like `n_neighbors`, `min_dist`, etc.  and a single value is specified in `aligned_umap.update` for the same parameter, `set_aligned_params` raises error. 
    - when the above bug is fixed, another hidden bug is revealed. `self.n_models_ += 1` update should be done after initializing the mapper in `aligned_umap.update`. Otherwise it returns `IndexError: list index out of range` error.

- Adds a test to reveal above bugs.
- Adds a test to check `ValueError` for the case a list of datasets is provided for `aligned_umap.update`.

### DIFF
--- a/umap/aligned_umap.py
+++ b/umap/aligned_umap.py
@@ -450,7 +450,6 @@ class AlignedUMAP(BaseEstimator):
         X = check_array(X)
 
         self.__dict__ = set_aligned_params(fit_params, self.__dict__, self.n_models_)
-        self.n_models_ += 1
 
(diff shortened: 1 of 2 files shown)


lmcinnes (approved):

lmcinnes:
Thanks, this is a great catch. I definitely appreciate the addition of the tests.

PR #896, opened 2022-08-05
Added uniqueness flag to intersect1d

SkBlaz:
Given the nature of the two arrays, this might offer consistently faster intersection computation.


lmcinnes:
Thanks.

PR #908, opened 2022-09-05
consistency between install_requires and requirements.txt

jlmelville:
My head hurts whenever I start reading about dependency management in python, but is the `install_requires` in `setup.py` intended to keep track with changes in `requirements.txt`? If so, here is a PR.


lmcinnes:
Every now and then I manage to read enough documentation that I can keep track of what the current way to do things is, and things get fixed to match, but mostly I don't really know. This seems sensible, so thanks.

Issue #848, opened 2022-03-30
Contents of precomputed_knn tuple


matthieuheitz:
I'm trying to feed to UMAP nearest neighbors computed with another method than pynndescent, and so I need to put the data in the format of precomputed_knn that UMAP accept: the tuple: (knn_indices, knn_distances, NNDescentObject)
Since I had trouble making an NNDescent object without giving it data, I tried giving it a fake NNDescent object, that returns true to isinstance(knn[2], NNDescent):

```python
import pynndescent
import numpy as np
import umap
import umap.plot
from umap.umap_ import nearest_neighbors
import matplotlib.pyplot as plt

n = 200
N = 3*n
y = np.random.rand(n, 60)
X = np.concatenate((y+20, y, y-20))
synthetic_labels = np.repeat([1, 2, 3], repeats=n)

random_seed = 10

knn = nearest_neighbors(
                        X,
                        n_neighbors=30,
                        metric='euclidean',
                        metric_kwds={},
                        angular=False,
                        random_state=random_seed,
                        )

knn_umap = umap.UMAP(n_neighbors=30, precomputed_knn=knn, random_state=random_seed, force_approximation_algorithm=True)
knn_umap.fit(X)


class MyNNDescent(pynndescent.NNDescent):
    def __init__(self):
        return

knn2 = (knn[0],knn[1],MyNNDescent())
knn_umap2 = umap.UMAP(n_neighbors=30, precomputed_knn=knn2, random_state=random_seed, force_approximation_algorithm=True)
knn_umap2.fit(X)

print("\033[1m"+"Are the embeddings for knn_umap and knn_umap2 the same?\033[0m")
print((knn_umap.embedding_ == knn_umap2.embedding_).all())

fig, ax = plt.subplots(1, 2, figsize=(13,7))
umap.plot.points(knn_umap, labels=synthetic_labels, ax=ax[0], theme='green')
umap.plot.points(knn_umap2, labels=synthetic_labels, ax=ax[1], theme='green')
ax[0].set_title("Precomputed knn 1st run", size=16)
ax[1].set_title("Precomputed knn 2nd run", size=16)
plt.show()
```

The result is that the 2 embeddings are the same, so it seems that the NNDescent object is not used at all.
Is my assumption correct, or is the object used in some particular cases, which would explain why it was included in the tuple?
I tried digging into the code, but I can't find the place where this object is used...

lmcinnes:
[...]

PR #909, opened 2022-09-05
Search index is optional for precomputed_knn

jlmelville:
This is a fix for #848 in the sense that it allows passing only indices and distances to as the `precomputed_knn` parameter (i.e. no pynndescent object required). Things to note:

* You will get a warning that you can't transform new data if you do this.
* If you then try to transform new data via the `transform` method, a `NotImplementedError` will be raised.
* If you have small data and `force_approximation_algorithm=False` you will *not* see the new warning, because the knn will not be used. Setting `force_approximation_algorithm=True` avoids this situation so this new feature works for big and small data.
* docstring for `precomputed_knn` parameter on the `UMAP` constructor has been updated.

Tests successes/warns/failures were the same before and after this addition (there is one failure currently which was failing on upstream-master for me before I added any code). I have added a new test to exercise these various cases.

### DIFF
--- a/umap/tests/test_umap_on_iris.py
+++ b/umap/tests/test_umap_on_iris.py
@@ -219,15 +219,6 @@ def test_umap_inverse_transform_on_iris(iris, iris_model):
         assert np.intersect1d(near_points, highd_near_points[0]).shape[0] >= 3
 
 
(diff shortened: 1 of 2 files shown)


lmcinnes:
Thanks, this looks like a nice simple way to resolve the issue.

PR #910, opened 2022-09-06
Fix doc build (a bit)

jlmelville:
Me again. Yes, I am on a tear today. 

When sphinx 1.8 came out (back in, er, 2018), the api changed `add_javascript` to `add_js_file`. I have updated the `conf.py` to reflect this and set a minimum requirement in `docs_requirements.txt` (although I can't imagine the latter being an actual problem).

Now I get a fair number of errors when I run `sphinx-build` but they are of the fixable type rather than of the oops-sphinx-exploded type.

Also I added a section to the `CONTRIBUTING.MD` about how to build the docs locally. I set the build output to that specified in `doc/Makefile` (which is `_build`) although I couldn't get `make` or `make help` to actually do anything.

Finally, I added the build output to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@ venv
 
 # metadata from pip-installing repo
 umap_learn.egg-info
+
+# docs
+doc/auto_examples
+doc/_build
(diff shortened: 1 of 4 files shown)


lmcinnes:
Thanks for this. I admit I haven't been paying attention to the doc builds recently, so noting that sphinx changed and it got broken is greatly appreciated.

PR #912, opened 2022-09-06
Allow points to be passed to umap.plot.points

dribnet:
Note sure if this change has drawbacks, but thought it might be worth considering and would easiest to discuss as a pull request. When I have large datasets I find it useful to do a train/test split and define the reducer with the training set, and then plot how it transforms the test set. But IIUC the API makes it tricky to then plot the result. This simple change allows umap.plot.points to plot new sets of points that come out of the reducer.

------8< commit message -----
The actual points to be plotted are usually
taken from the reducer returned by umap.UMAP.
However it is possible to generate other sets
of points by subsequently calling reducer.transform().

This change allows the plot.points call to
optionally accept the points array so that
other sets of points can easily be plotted
other than those that defined the reducer.

### DIFF
--- a/umap/plot.py
+++ b/umap/plot.py
@@ -480,6 +480,7 @@ def show(plot_to_show):
 
 def points(
     umap_object,
+    points=None,
     labels=None,
     values=None,
     theme=None,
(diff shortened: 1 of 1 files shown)


lmcinnes:
I'm sure this will come in handy. Thanks for taking the time to contribute it back.

PR #913, opened 2022-09-06
Fix the doc build

jlmelville:
This PR makes `sphinx-build -b html doc doc/_build` complete without warnings or errors. What this required:

* modifying the indentation and other small formatting changes to the docstring and a couple of the `.rst` docs.
* I excluded the UMAP class definition from appearing in the API docs twice (it was appearing once as itself, and once as a module).
* I needed to install `numpydoc` to make the `Parameters`-style docstrings not complain. Possibly `napoleon` was meant based on discussion online, but I don't know enough about this (it did also require adding an extra directive in `conf.py` to stop a complaint about stub files).
* A lot of the sphinx gallery reference URLs to other packages (e.g. Pandas) no longer seem to offer a correct `search.js`, so have been commented out.

I haven't actually checked the generated documents (the penguins dataset example definitely needs some URLs fixed) but this is a good start I think.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -16,4 +16,5 @@ Useful Functions
 
 .. automodule:: umap.umap_
    :members:
+   :exclude-members: UMAP
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -234,5 +234,5 @@ def setup(app):
 
 
 # This prevents complaints autosummary stub files not being found for each method on
-# the class. See https://stackoverflow.com/questions/65198998/sphinx-warning-autosummary-stub-file-not-found-for-the-methods-of-the-class-c
+# the class. See https://stackoverflow.com/q/65198998
 numpydoc_show_class_members = False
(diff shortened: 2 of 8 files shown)


lmcinnes:
Thanks for working through all of this. Hopefully it all works now ...

jlmelville:
Looks like the doc build still fails. I don't see a `.readthedocs.yaml` file, so I assume the docs are built via a readthedocs.org admin panel?

lmcinnes:
It should happen automatically; I'm pretty busy at the moment, but when I get some time I'll see if I can fix this.

lmcinnes:
Looks like it was numpydoc having to be added to doc_requirements. I think it has run correctly now.

jlmelville:
Yeah I see the doc build green at last. This gets a 🥳 from me. When I look at <https://umap-learn.readthedocs.io/en/latest/api.html#umap> I don't see the bullet points for the `metric` section when the metrics are listed, but I *do* see it locally. Hopefully that's the worst of it though.

To celebrate I have a PR to fix the broken `penguins` links (and update to the changed format) in the how-to-use-umap doc forthcoming. Nothing urgent though.

PR #914, opened 2022-09-16
Update penguins doc

jlmelville:
Striking while the doc-build iron is hot, this PR:

* updates some URLs that had got stale. In particular, the URL for the penguin image, but also the CSV file itself.
* the URLs are now github permalinks, so hopefully even if future breaking changes occur, these particularly URLs will stay valid.
* updates to the penguin dataset meant that:
    * there are now a different number of non-NA rows.
    * also the column names had changed slightly.
    * *and* there was a `year` column added.
* the text has been adjusted to reflect the above and I drop the `year` column in the pairplot.
    * I had to regenerate the pairplot. It has larger dimensions (not sure how to change that via jupyter lab) but is otherwise quite similar.
* also to be on the safe side I regenerated the UMAP plot (it doesn't change much, reassuringly).
* one change to the digits example: there was a minor change to the `PIL` API.


jlmelville:
how did I break all the builds with this change? 🤔

lmcinnes:
The breakage is, I think, my fault. Don't worry about it for now.

PR #924, opened 2022-10-08
new spectral initialization option: a truncated SVD-warmed lobpcg 

jlmelville:
As mentioned in #918 I have noticed that some datasets cause the spectral initialization step to take a lot longer than usual. These include several used in the [Stochastic Cluster Embedding paper](https://arxiv.org/abs/2108.08003), and which are available for download from the [sce github repo](https://github.com/rozyangno/sce). In particular [tomoradar](https://github.com/rozyangno/sce#tomoradar-data-set), [flow cytometry](https://github.com/rozyangno/sce#flow-cytometry-data-set) and [shuttle](https://github.com/rozyangno/sce#shuttle-data-set) take longer than they should (shuttle isn't so bad in absolute terms, but the other two take several minutes). I have also noticed that the [mammoth](https://pair-code.github.io/understanding-umap/) dataset in the form used by Understanding UMAP takes a surprisingly long time.

As a solution I propose that [`lobpcg`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.lobpcg.html), which is only used for very large problems in the `spectral_layout` function, can work quite well as a general alternative to [`eigsh`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.eigsh.html) *if* you provide a decent initial guess for the eigenvectors. And it turns out [truncated SVD](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html) can provide that guess: you can shift the graph laplacian cheaply and the singular vectors will correspond to the unshifted laplacian's eigenvectors. The settings `algorithm="arpack", tol=1e-2` work well.

With these changes (and with `n_neighbors=50` to avoid complications from more than one connected component in the input graph) I was able to get `cytometry`'s initialization to drop from 20 minutes to 2 minutes, `tomoradar` from 4 minutes to 20 seconds, `shuttle` from 54 seconds to 8 seconds, and `mammoth` from 50 seconds to 8 seconds, which seems like a worthwhile time saving. The layouts that result seem to be at least as reasonable as those with `spectral_layout`.

I also looked at datasets that were not having problems with `spectral_layout` and while these do not run consistently faster with the new method, none of them suddenly started taking anomalously long either. UMAP results also looked in line with what I expect. That said, this is not supposed to *replace* the existing `spectral_layout`, but to provide an alternative for difficult cases. I have also only looked at `n_components=2`, but I assume that is the most common use case with UMAP.

## The implementation

* I have added this as a separate function `tsw_spectral_layout`. TSW as in "Truncated Svd Warmed", because it provides a "warm" start to `lobpcg`. Horrid name I know. I am open to suggestions for something better.
* You can get it by setting `init="tswspectral"`.
* As it is a bit experimental it does not replace or insert itself into any other part of the initialization. If multiple components are detected in the input graph, it bails out to `multi_component_layout` just like `spectral_layout`.
* I envisage it being used as an initialization of last resort if the standard default spectral layout is taking way too long and seems to have complete ground to a halt.
* If convergence fails then you get a warning rather than an error being raised, so we look for that. That means other warnings get caught in the dragnet, but whatever calls the warning, it's rarely good news for the result so falling back to random initialization in any case is probably justifiable.
* I have added a test that the result of `tsw_spectral_layout` and `spectral_layout` are close.
* for some reason, when I ran `black` other bits of the modules got reformatted. Not sure how they got away for so long. I've kept those changes in this PR.

## Some obvious questions

This bit is only worth reading if you got this far and thought "this seems a bit ornate, surely there's an easier way to make `spectral_layout` work better?!". That may be the case, imaginary interlocutor, but I was unable to find it:

* can the `eigsh` routine be improved? `eigsh` doesn't let you provide a guess for all the eigenvectors, just the first one (`v0`). Now as it happens there may be some confusion around the eigenvectors of the graph laplacian in the code because currently we pass a vector of ones as the guess. However, while that is the smallest eigenvector associated with the un-normalized graph Laplacian, the smallest eigenvector for the symmetrized normalized graph laplacian is actually $D^{1/2}$ which we already calculate (just take the square root of `diag_data[0]`). Infuriatingly while this does often help, it doesn't consistently do better than guessing all ones for `v0`. Oh well.
* `eigsh` uses arpack under the hood just like truncated SVD, why not use that? I was not able to get `eigsh` to produce decent cheap results as easily as truncated SVD.
* or what about using the `shift` features of either `eigsh` or `lobpcg` and use the shifted laplacian with them instead of truncated SVD? Whenever I tried this it actually made things a lot slower.
* can we improve the guess to `lobpcg` without getting truncated SVD involved, by for example passing $D^{1/2}$ as the first column? Unlike `eigsh`, that does actually help and reduced the number of datasets where convergence failed, but there were still some failures, while convergence was always seen with the truncated SVD guess.
* is there a cheap diagonal preconditioner available for use with `lobpcg`? My understanding is that the inverse of a graph laplacian is related to the commute distance between items in the dataset but I couldn't find a cheap and easy way to approximate that. 
* is the truncated SVD routine on its own good enough to provide the eigenvectors directly? In my tests, it is not competitive with `eigsh` and `lobpcg` at the tolerances we require.
* instead of the `arpack` algorithm for truncated SVD, what about the `randomized` version? That algorithm is fast, but seems to give a good low-rank approximation, but not necessarily a highly accurate SVD, which `lobpcg` really needs (at least so it would seem for UMAP-created graph laplacians).

### DIFF
--- a/umap/spectral.py
+++ b/umap/spectral.py
@@ -133,7 +133,7 @@ def component_layout(
                     component_centroids, metric=metric, **metric_kwds
                 )
 
-    affinity_matrix = np.exp(-(distance_matrix ** 2))
+    affinity_matrix = np.exp(-(distance_matrix**2))
 
     component_embedding = SpectralEmbedding(
         n_components=dim, affinity="precomputed", random_state=random_state
--- a/umap/spectral.py
+++ b/umap/spectral.py
@@ -1,9 +1,12 @@
+import warnings
+
 from warnings import warn
 
 import numpy as np
 
 import scipy.sparse
 import scipy.sparse.csgraph
+import sklearn.decomposition
 
 from sklearn.manifold import SpectralEmbedding
 from sklearn.metrics import pairwise_distances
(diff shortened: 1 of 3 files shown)


lmcinnes:
Thanks for this. I have in the past run up against the spectral layout having odd issues, and had as much (or slightly more, since I never came up with a solution like this) difficulty as you. This makes a lot of sense as far as I can tell. I may actually end up adding a deprecation notice and swapping this in as the standard spectral layout in due course. Thanks for all the work figuring this out.

Issue #926, opened 2022-10-09
UMAP based on correlation matrix


bigsuicideparty:
I would like to use UMAP as a preprocessing step for density based clustering (DBScan) of a multidimensional data set. Is it possible to utilize a correlation matrix that I have generated for my data in this approach? According to the [[manual](https://umap-learn.readthedocs.io)][1] it should be possible to provide a precomputed k-nn. I was thinking of converting my correlation matrix to a distance matrix and use this as an input for the nearest neighbor algorithm. Do you think this is a sensible approach? If so, has anyone of you suggestions on how to properly implement this in python? My thoughts were something along the line of:

starting with a correlation matrix (*correlation_matrix*) of my actual data matrix (*df_data*)

   ```
 import umap.umap_ as umap
    from umap.umap_ import nearest_neighbors

    dist_matrix = 1 - correlation_matrix
    knn_correlation = nearest_neighbors(dist_matrix,
                              n_neighbors=15,
                              metric="precomputed",
                              metric_kwds=None,
                              angular=False,
                              random_state=None,
                             )

    knn_embeddings = umap.UMAP(n_neighbors=15,min_dist=0.1,
                        precomputed_knn=knn_correlation,
                        ).fit_transform(df_data)
```

 
Has any come across a similar situation or has any suggestions? Thank you!

  [1]: https://umap-learn.readthedocs.io

**Comment by jlmelville** [comment]
What you have is essentially correct. There is one (maybe two) issues:

1. you need to persuade UMAP to accept pre-computed nearest neighbors, which you can do by replacing the third element of the tuple returned by `nearest_neighbors` with a dummy class.
2. if your dataset is small, you need to set `force_approximation_algorithm=True` when you create the `umap.UMAP` object. 

To remedy this, insert something like this after creating `knn_correlation` and before calling the `umap.UMAP` constructor:

```python
import pynndescent 


class DummyNNDescent(pynndescent.NNDescent):
    def __init__(self):
        return

knn_correlation = (knn_correlation[0], knn_correlation[1], DummyNNDescent())
```

Then where you are calling `umap.UMAP(n_neighbors=15, min_dist=0.1, precomputed_knn=knn_correlation)` also add `force_approximation_algorithm=True` as an argument.

[...]

PR #927, opened 2022-10-10
ignore force approximation with precomputed knn

jlmelville:
In #909 I added the ability to provide a precomputed knn without also having to fake out UMAP with a dummy NNDescent class. But you still had to explicitly set `force_approximation_algorithm=True` in the small data case, which is odd and confusing. In #848 @lmcinnes commented that this should probably be an allowed scenario.

I am revisiting the issue due to #926, where a pre-computed knn came up again and I realized that even when #909 becomes part of an official UMAP release, it remains awkward to have to remember to  set `force_approximation_algorithm=True`. 

This PR now forces `force_approximation_algorithm=True` when the data is small and precomputed knn is provided, ignoring what the user provides (note that this parameter was only added to help with unit testing and is not documented). A simple example of creating a distance matrix and then using it with UMAP now looks like:

```python
import matplotlib.pyplot as plt
import sklearn.metrics
import umap.umap_ as umap
from sklearn.datasets import load_iris
from umap.umap_ import nearest_neighbors

iris = load_iris()
iris_dm = sklearn.metrics.pairwise_distances(iris.data)

knn_correlation = nearest_neighbors(
    iris_dm,
    n_neighbors=15,
    metric="precomputed",
    metric_kwds=None,
    angular=False,
    random_state=None,
)

knn_embeddings = umap.UMAP(
    n_neighbors=15,
    min_dist=0.1,
    precomputed_knn=knn_correlation,
).fit_transform(iris.data)

plt.scatter(knn_embeddings[:, 0], knn_embeddings[:, 1], c=iris.target, alpha=0.5)
```

You will still get a warning that the umap transformer cannot be used for transforming new data (which is correct), but there is no longer the need to set any unexpected parameters, nor is there the risk that your knn will be ignored if you do forget to set an undocumented parameter, so IMO this represents a substantial improvement over the same example in #926.

The only other code changes are for the tests which I originally added for #909 which characterized when warnings would and wouldn't occur. This change simplifies that logic substantially so the number of tests has been reduced.

### DIFF
--- a/umap/tests/test_umap_on_iris.py
+++ b/umap/tests/test_umap_on_iris.py
@@ -243,65 +243,42 @@ def test_precomputed_knn_on_iris(iris, iris_selection, iris_subset_model):
         min_dist=iris_subset_model.min_dist,
     )
 
-    # Trying to pass a precomputed knn with a small dataset gets a warning and the
-    # precomputed knn is ignored
-    with pytest.warns(Warning, match="your data is small") as record:
-        fitter_with_knn_ignored = UMAP(
-            **umap_args,
-            precomputed_knn=knn,
-        ).fit(data)
-        assert len(record) == 1
-    # you therefore get the same output as the iris subset model result
-    np.testing.assert_array_equal(
-        fitter_with_knn_ignored.embedding_, iris_subset_model.embedding_
-    )
+    # force_approximation_algorithm parameter is ignored when a precomputed knn is used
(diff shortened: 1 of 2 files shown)


lmcinnes:
Thanks for this; I think it cleans up the issue in a nice way. I am hoping to get a release out for various projects (hdbscan needs one as well) when I get back from vacation in a few weeks.

PR #929, opened 2022-10-11
Update deprecated matplotlib colormap registry

yannikschaelte:
`matplotlib.colormaps.register_cmap` is deprecated in favor of `matplotlib.colormaps.register` since 3.6.0, see https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/cm.py#L235.


lmcinnes:
Thanks!

PR #936, opened 2022-11-02
Esm metagenomic atlas to Interactive Visualizations

BioGeek:
Adding the [ESM Metagenomic Atlas](https://esmatlas.com/) to the Interactive Visualizations. the text is taken verbatim from their [About](https://esmatlas.com/about) page. 

More technical details can be found in the [preprint](https://www.biorxiv.org/content/10.1101/2022.07.20.500902v2.full.pdf) (pdf):

>  To construct the landscape of MGnify sequences, we first used ESM-1b to embed each sequence as a 1280-dimensional vector. These embeddings were then visualized using the umap version 0.5.3, scanpy version 1.9.1, and anndata 0.8.0 Python packages, where dimensionality reduction was applied directly to the embedding vectors
(use rep=‘X’ in scanpy.tl.umap) with default parameters (15-nearest-neighbors graph via approximate Euclidean distance, UMAP min dist=0.5). 

I have no relation to the authors of the paper, just came across this tweet: https://twitter.com/BrianHie/status/1587478389841006592


lmcinnes:
Thanks -- this looks like an amazing project!

Issue #938, opened 2022-11-11
[FEATURE] `get_feature_names_out` for `UMAP`


CarloLepelaars:
I propose to add a `get_feature_names_out` method to the `UMAP` object.

`get_feature_names_out` is an important component for interpreting scikit-learn Pipeline objects. A `get_feature_names_out` call on a `Pipeline` only works if it is implemented for all components in the pipeline, except the last step (i.e. the Model).

Scikit-learn recently implemented `get_feature_names_out` for all Transformers in their 1.1 release ([Source](https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_1_0.html?highlight=scikit+learn+release#get-feature-names-out-available-in-all-transformers)).

Example implementation:
```python
class UMAP(BaseEstimator):
    .
    .
    .
    def get_feature_names_out(self, feature_names_out=None):
        return ['umap_component_1', 'umap_component_2']
```

lmcinnes:
I would be happy to include it if you want to make a PR.

CarloLepelaars:
👍 I've made a PR which implements this and run multiple tests. Let me know what you think!

PR:
https://github.com/lmcinnes/umap/pull/939

PR #939, opened 2022-11-11
`get_feature_names_out` for `UMAP` + tests

CarloLepelaars:
Solves issue #938


lmcinnes:
Looks great. Thanks for this --  as you noted it will be very helpful for pipelines.

PR #977, opened 2023-03-01
Allow UMAP to accept data with nan or inf values

SamKlier:
Based on our [discussion](https://github.com/lmcinnes/umap/discussions/953) about nan-values and custom metrics, I finally had time to (hopefully) properly address the issue.

I exposed the 'force_all_finite' parameters of sklearn's check array functions to UMAP's data handling functions. The default is "force_all_finite=True" which means that this change shouldn't break anything. I also included a test case to test data inputs in respect to accepting/dismising nan's or inf's.

However, I DON'T check if the selected metric is fine with nan or inf values!

### DIFF
--- a/.idea/umap-nan.iml
+++ b/.idea/umap-nan.iml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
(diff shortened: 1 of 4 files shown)


lmcinnes:
Sorry for taking so long to get to this. Thanks!

SamKlier:
Don't worry ;)

PR #993, opened 2023-04-05
Changes in Spectral Layout

Sakshi-2797:
This contribution includes a faster implementation of the spectral_layout function. The implementation replaces the eigenvalue compute with a faster MKL based method. The modification provides a  speedup of **~9x** in some cases.


lmcinnes:
While I appreciate the work, and the speedup would be helpful, I am wary of a new dependency on a binary library like this -- I also note you have an environment variable set that is hardcoded to a path that may not be the install location on some machines, and won't work at all on non-Linux machines.

Sakshi-2797:
> While I appreciate the work, and the speedup would be helpful, I am wary of a new dependency on a binary library like this -- I also note you have an environment variable set that is hardcoded to a path that may not be the install location on some machines, and won't work at all on non-Linux machines.

Hi @lmcinnes ,
We have made the required changes , the environment variable is changed to a dynamic implementation which will work on the on Linux and non-Linux machines. However , we are facing some issues while raising a PR as some pre commit tests aren't passing. Is there a way to run these tests on local environment?

PR #1000, opened 2023-04-24
Changes in Layouts , Addition of new files

Sakshi-2797:
This contribution includes an alternate cpp implementation of the "optimize_layout_euclidean" function in layouts.py. The changes includes a faster cpp implementation of the optimization function by introducing separate files - umap_extend.cpp, xsrandom.cpp,xsrandom.hpp.

### DIFF
--- a/umap/layouts.py
+++ b/umap/layouts.py
@@ -1,6 +1,7 @@
 import numpy as np
 import numba
 import umap.distances as dist
+import umap_extend
 from umap.utils import tau_rand_int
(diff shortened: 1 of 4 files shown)


lmcinnes:
This looks like impressive work, but for my own sake I would like to keep this version of umap as a python only library (for ease of maintenance and packaging). I would encourage you to build this work into a fork that can be a package you maintain (``fast_umap`` perhaps?). You could roll in your spectral improvements there as well, and other speedups as you see fit.

Issue #1007, opened 2023-05-12
Latest umap-learn 0.5.3 is not compatible with latest numpy 1.24.3


orenbenkiki:
Umap uses `np.int` in `umap/umap_.py` line 2465 and `np.float` in line 2467. However these have been removed so the code fails. The error messages suggested replacing them with a simple `int` and `float`. This fix worked for me.

orenbenkiki:
Noting this isn't fixed yet and causes umap to fail when running in modern Python installations.

PR #1010, opened 2023-05-22
fix: fix numpy.int and numpy.float deprecation to be compatible with numpy >= 1.20.0

TTTPOB:
Since 1.20.0, numpy has marked np.int and np.float as [deprecated](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations), and since 1.24.0, np.int has been [expired](https://numpy.org/devdocs/release/1.24.0-notes.html#expired-deprecations)
Numpy has suggested use `int` to replace `np.int`

related: #1007


lmcinnes:
Thanks!

PR #1019, opened 2023-06-07
Fix docstring formatting of init parameter of UMAP class

felixdivo:
The docsting is a bit hard to read without this:

Screenshot 2023-06-07 at 10 54 13


felixdivo:
I'm not certain that it is fixed now, one should check that first.

lmcinnes:
Thanks!

PR #1045, opened 2023-08-21
Drop dependency on TBB on non-Intel deployments

hamelin:
The project depends on package `tbb', short for Intel's Thread Building Blocks. Obviously, this cannot possibly install on to ARM processors. This PR arranges for the setup script to include the dependency only on Intel-based machines (regardless of OS).


hamelin:
This PR depends on #1044 let's clear that one first.

PR #1050, opened 2023-09-01
Code refactoring 

tosemml:
The objective of these minor code modifications is to improve code conciseness through the utilization of highly optimized NumPy operations.


lmcinnes:
These operations are in numba compiled functions, so loops are not actually slow, and can often be faster than numpy. Can you provide benchmarks demonstrating that this is actually any faster?

PR #1053, opened 2023-09-08
Add azure devops pipeline to push to PyPI

gclendenning:


hamelin:
Some changes posterior to unit-testing commits discarded the `requirements.txt` file, in favor of installing locally using the `setup.py` script instead. This removes duplication of the package dependencies. Can you revise your sync so as to avoid overwriting this change?

gclendenning:
I took out the requirements.txt parts. Some of the test jobs failed due to time-out but I think it should be good

hamelin:
@lmcinnes this looks good to me.

PR #1061, opened 2023-10-05
Fix numba call for deprecation of nopython=True

### LINKED ISSUE #1048: Numba.jit and numpy dependency problems
Hi, I'm getting this warning with umap 0.5.3:

```
  @numba.jit()
/Users/kwaneu/sw/miniconda/miniconda3/envs/python/lib/python3.11/site-packages/umap/umap_.py:660: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.
```

I'm also seeing umap complain it only supports up to numpy 1.24.0.

aarmey:
Fixes #1048 for umap.plot.

I think this line was missed during the changes in v0.5.4.

### DIFF
--- a/umap/plot.py
+++ b/umap/plot.py
@@ -200,7 +200,7 @@ def _nhood_search(umap_object, nhood_size):
     return indices, dists
 
 
-@numba.jit(nopython=False)
+@numba.njit()
 def _nhood_compare(indices_left, indices_right):
     """Compute Jaccard index of two neighborhoods"""
     result = np.empty(indices_left.shape[0])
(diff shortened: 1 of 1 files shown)


lmcinnes:
This is actually a nopython=False case; I think in practice it probably needs to be updated to use the python object context handling that has been added to numba. I'll try to get around to that in due course.

aarmey:
@lmcinnes you might want to avoid the additional complexity, but here is a version that can be fully compiled by Numba.

PR #1072, opened 2023-11-14
Fix `pkg_resources` deprecation warning

laclouis5:
The package `pkg_resources` is deprecated. This PR implements using `importlib` instead for getting the version number.


laclouis5:
Note that the tests fail because Python3.7 is tested and `importlib` was only added in Python3.8. However, Python3.7 has reached end-of-life and won't be maintained in the future. I think that it should be a reasonable choice to drop support for Python3.7.

lmcinnes:
Thanks!

Issue #631, opened 2021-03-27
umap takes 10 seconds to import


jdkern11:
First off, great work with umap, it's a very cool tool. 

I am trying to make a GUI that performs some mapping for a user. My issue is, it takes 10 seconds to import umap, which means the GUI start up is very slow. No other dependencies take this long, so I was wondering if there is some way to speed it up?

I have other features in the GUI, so ideally the user could perform other actions while the umap is created with cross threading. It's just the slow import that concerns me.

lmcinnes:
I'm not actually sure what takes up time in the import phase. At least some of it may be related to numba compilation, but it could be any number of things. I can't offer any immediate solutions I am afraid.

jdkern11:
That's okay, I found I could import/run it in a PySide2 worker thread in the background and the user can still operate the GUI while it's loading. I appreciate the quick response however!

zapr-in:
It seems most of the time is taken by pynndescent
```shell_session
$ time python -c "import umap"
real	0m13.641s
user	0m13.729s
sys	0m0.518s

$ time python -c "import pynndescent"
real	0m13.315s
user	0m13.435s
sys	0m0.469s

```

rsuhada:
Sorry to comment on this closed issue, but is there a way to speed up the import?
The PySide2 workaround is not applicable to my case.

UMAP gives very good results for my application and embedding calculation fast enough. However, we can't put it into production because the import is exceedingly long (cloud deployment).

Thank you!

jdkern11:
Is the user uploading data that will be incorporated into the map? If not, you could premake the map and save the embedding, then just load the embedding to display.

rsuhada:
The embedding has to be calculated for each user/session separately, so we can't use a cached one (and due to the system architecture, the library has to be imported at request time). 
The users are non-technical and can't wait 10 seconds for the results...

At the moment I've reverted to Kernel PCA (which is as expected not as good, but usable).

jdkern11:
Can you add a loading bar to the interface? 10 seconds is slow but not so slow that most people would just leave if you have a loading bar.

PR #1090, opened 2024-02-01
Import pynndescent lazily to speed up import

sliedes:
Importing umap tended to take long; pynndescent is the culprit. This moves the substantial cost of importing pynndescent from import time to the first use. Among other things, this is likely to make it much nicer to run `script.py --help` and other operations that actually do not use UMAP on programs that nevertheless import UMAP.

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -45,9 +45,34 @@ from umap.layouts import (
     optimize_layout_inverse,
 )
 
-from pynndescent import NNDescent
-from pynndescent.distances import named_distances as pynn_named_distances
-from pynndescent.sparse import sparse_named_distances as pynn_sparse_named_distances
+
+class PynndescentWrapper:
+    def __init__(self):
+        self._module = None
+
+    @property
+    def _get(self):
+        if self._module is None:
+            import pynndescent
+
+            self._module = pynndescent
+        return self._module
+
+    @property
+    def NNDescent(self):
+        return self._get.NNDescent
+
+    @property
+    def named_distances(self):
+        return self._get.distances.named_distances
+
+    @property
+    def sparse_named_distances(self):
+        return self._get.sparse.named_distances
+
+
+pynn = PynndescentWrapper()
+
 
 locale.setlocale(locale.LC_NUMERIC, "C")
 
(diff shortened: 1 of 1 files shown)


sliedes:
Related old issue: #631

lmcinnes:
I understand that import time can be annoying. On the other hand I'm not convinced that punting it to first use necessarily improves the experience. Long term I would like to have caching work better in numba and then pynndescent will take less time to load (as we can safely cache more) but that's a different problem.

I'm not entirely averse to this proposal, but would it be possible to pitch me on why you think this approach will be better?

sliedes:
Thank you for giving me the opportunity to pitch!

It's likely not uncommon for libraries using UMAP either conditionally or unconditionally to import it in a module; generally, it requires work (similar to the patch I proposed here) to do things more lazily. For example, I come from using BERTopic, which by default uses UMAP when fitting a model, but I'm not sure it needs it when just loading and using a model.

I view it as more of an interactivity and iteration time issue. When I execute a script that indirectly imports UMAP, I only get to see any trivial error messages before getting to the point where UMAP would be used after several seconds due to the import time. Just running `python script.py --help` takes several seconds.

With this change, I'd assume most such scripts would be able to perform the error-prone steps of command line argument parsing, data loading etc. without having to spend the time it takes to import pynndescent.

In the case where UMAP is actually not needed (like I think with BERTopic when loading a model), the execution time of the import is, of course, saved entirely.

Now, it would obviously be possible for all such modules to do the same thing to lazily import UMAP, but then each of them would need to do it separately. Generally, I agree with you that the ideal case would be fixing things in numba so that pynndescent can be loaded faster; failing that, pushing the laziness as deep in the chain as possible results in the most generally useful fix in the sense that every user of it will automatically get the benefits. That is, the second nicest thing would be if pynndescent could only do its compilation lazily, but I understand that due to numba limitations this is not currently feasible without, I think, a more involved wrapping there. Having said that, if you wish, I can try to develop a similar workaround for pynndescent.

lmcinnes:
You do make a decent case. On the other hand it will result in a significantly slower first run of UMAP because it will be deferring the import to that point, and as that import is hidden the delay won't be obvious to users. Let me think about it for a bit.

Issue #580, opened 2021-02-10
[RFC] Feature: `ParametricUMAP` PyTorch implementation


leriomaggio:
In the spirit of the PR #578 which allows dependencies not to interfere too much with the test suite, I was wondering whether you would consider adding an extra dependency 😅 

More seriously, I started working to an alternative PyTorch-based implementation for `ParametricUMAP` (will soon sync the branch on my fork)

If you would generally appreciate the idea and the contribution, I will elaborate better here my UEP (_UMAP Enhancement Proposal_) to have the two implementations co-existing, with the less impact possible in the API for backward compatibility.

lmcinnes:
I would like to get @timsainb 's opinion on this. Personally I can see a lot of value in supporting other deep learning frameworks, and since dependencies can be optional I don't see and extra potential dependency as a problem. I do know that other have asked about a PyTorch version, so there is certainly interest. I think it would be a very exciting addition. The key would be to ensure it syncs up with @timsainb 's implementation in TF and doesn't cause (too many) weird interactions as to what happens under the different backends. It sounds like you have thought about that, so please feel free to elaborate on that -- I would like to hear more.

timsainb:
I think that would be a great addition.

lmcinnes:
If Tim is on board, then I think we can all this a go! Really looking forward to seeing this in place.

leriomaggio:
Fantastic!! 🙌  
Thanks so much @timsainb and @lmcinnes for getting back to me about this! 
Will submit my initial PR very soon so that we can start discussing design options there. 

I will be sharing quick ideas here as well while I do work on it, so that we could potentially upvote/downvote options very quickly - will be starting soon in my next post!

Thank you so much :)

leriomaggio:
**Q**: How do we allow multiple DL `backend` for Parametric UMAP (so far, `tf` and `torch`)

_iow_, what implementation will I import whenever I will do: 
```python 
from umap.parametric_umap import ParametricUMAP
```
My idea would be in favour of a _Convention of Configuration_ based approach - so **no** configuration file `a-la-keras` to choose the backend

**General Idea**

- If `torch` is available but not `tf` --> go with `torch`
- if `tf` is available but not `torch` --> got with `tf`
- if `tf` and `torch` are both installed in the environment --> go by default with @timsainb / `tf` implementation (unless specified - as in importing the default torch-based class/implementation)

[...]

PR #1103, opened 2024-03-23
Torch implementation of ParametricUMAP

jh83775:
Hi All,
I've popped in a pull request with a version of ParametricUMAP written in PyTorch. It mostly follows the discussions in the threads below:
* https://github.com/lmcinnes/umap/issues/580
* https://colab.research.google.com/drive/1CYxt0GD-Y2zPMOnJIXJWsAhr0LdqI0R6

I'm not quite sure where it best fits in the codebase - currently I've just put it in umap/torch.py so you can import with `from umap.torch import ParametricUMAP`, which feels intuitive. It should be fairly easy to edit and include a nonparametric UMAP as well, via the `torch.nn.Embedding` class, but I can put in a fresh pull request for that when I find the time. There could also be a discussion around what parameters are accessible to the user while not bloating the number of optional args. For example, in my implementation the user can manually change the batch size and learning rate, but not the optimizer or negative sample rate. What is included in the optional args is relatively arbitrary based on what I tended to be changing the most for my use case. It will be relatively easy to expose any currently hardcoded parameter choices to the user though.

I've also added a few examples in a Jupyter notebook (see `notebooks/Parametric_UMAP/08.0-torch-parametric-umap.ipynb`) to give an indication of how to get started.

If any PyTorch experts want to take a look, I'll gladly add in any optimizations (I'm sure there will be a few things here and there that can be changed and improve the runtime).

I've tested it with pytorch==1.12.1 & CUDA==11.6. I think it should work with other versions too, but haven't tested more widely.


lmcinnes:
Thanks for this. There is currently work underway to support Keras 3 (the initial support is done) such that multiple backends, including pytorch, could be used. You may want to check in on the refactoring discussion on PR #1101

jh83775:
> Thanks for this. There is currently work underway to support Keras 3 (the initial support is done) such that multiple backends, including pytorch, could be used. You may want to check in on the refactoring discussion on PR #1101

That looks helpful, thanks. I'll update to reflect the discussions on that thread. Looks like the keras-pytorch version might supersede this work, although hopefully this can still be useful for those more familiar with torch than keras

lmcinnes:
I think there's still merit in this; I just don't have a good sense of
where things will settle to know what the right way to support the various
options will be.

lmcinnes:
So as far as I can tell this will just sit alongside the existing Keras3 implementation (which should now support a pytorch backed) just fine as an extra pytorch specific ParametricUMAP. I would be happy to just merge this if you think it has sufficient value as-is for a standalone pytorch speciifc implementation.

timsainb:
> So as far as I can tell this will just sit alongside the existing Keras3 implementation (which should now support a pytorch backed) just fine as an extra pytorch specific ParametricUMAP. I would be happy to just merge this if you think it has sufficient value as-is for a standalone pytorch speciifc implementation.

I'm also glad to see this implementation. Thanks! 

I do think the overlap in functionality between this and the keras implementation is pretty high. 

This code is missing some key functionality (it's been a few days since I read it). One of the benefits of parametric UMAP is it allows you to balance the global MDS loss and auto encoder loss, which iirc are not in this implementation. 

One thing that the keras code is missing that this code has is the data iterator. Currently to run the keras code, even if you use torch as a backend you still rely on on the tensorflow dataset, meaning you need to have keras, tensorflow, and torch installed. If we pull the iterator from this code into the keras implementation we will overcome that issue (this is something I've been meaning to do but haven't had time yet)

turian:
@timsainb as per discussion on [HN](https://news.ycombinator.com/item?id=40310051) by Max Woolf, one big limitation of the current parametric UMAP implementation is OOM. Caused by tensorflow dataset pulling the entire graph into memory.

So a parametric UMAP implementation that works iterating over the dataset would be great.

timsainb:
@turian I think this is out of the scope of the umap python library, but I give an example of how to build your own graph and iterator here: https://colab.research.google.com/drive/1WkXVZ5pnMrm17m0YgmtoNjM_XHdnE5Vp?usp=sharing

You could switch out the graph and the interator for one that supports a graph larger than memory and an iterator that grabs from that graph.

PR #1126, opened 2024-05-27
Add the option to exclude raw data from the saved model

bartbroere:
This relates to the discussion in #1118 

The approach I chose to do this was removing the raw data from the object, and restoring it after saving. This way the `save` method has no unintended side effects.

If this will be merged, or something similar will be added, I'm willing to contribute some documentation as well.


timsainb:
We will need to merge this with https://github.com/lmcinnes/umap/pull/1123

bartbroere:
> We will need to merge this with #1123

I could target that branch with this PR, if that helps.

PR #1147, opened 2024-08-13
Update parametric_umap.py

AMS-Hippo:
Added short functions for exporting the encoder to torch or ONNX formats.

### DIFF
--- a/umap/parametric_umap.py
+++ b/umap/parametric_umap.py
@@ -358,8 +358,8 @@ class ParametricUMAP(UMAP):
         # Extract encoder
         km = self.encoder
         # Extract weights
-        pm = PumapNet(self.dims[0],self.n_components)
-        pm = to_torch(km,pm)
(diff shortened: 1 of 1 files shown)


umap/parametric_umap.py
```diff
@@ -1019,3 +1044,72 @@ def _parametric_reconstruction_loss(self, y, y_pred):
         loss = self.parametric_reconstruction_loss_fn(
             y["reconstruction"], y_pred["reconstruction"])
         return loss * self.parametric_reconstruction_loss_weight
+
+##################################################
+# 1. Pytorch version of parametric UMAP network. #
+##################################################
+
+
+class PumapNet(nn.Module):
```
lmcinnes, line 1053: I think you need to wrap this class definition in a conditional on torch.nn having been imported -- it is currently causing the tests to fail in the CI.

lmcinnes:
Thanks!

PR #1152, opened 2024-10-11
Compile _optimize_layout_euclidean_single_epoch once

kmkolasinski:
Hi this is a simple fix which makes `optimize_layout_euclidean_single_epoch` to be compiled once per python process life time. 

## Explanation
In some cases we need to run UMAP thousands of times on datasets of order 100-1000 points. Since these datasets are very small umap  usually needs a fraction of second to complete, but because we have to `numba.njit` each time we have to wait for the result approximately 0.5 second. 

In our case we have to run umap let's say 10k times, which gives about extra 80 minutes to such process.

### DIFF
--- a/umap/layouts.py
+++ b/umap/layouts.py
@@ -219,6 +219,22 @@ def _optimize_layout_euclidean_densmap_epoch_init(
         re_sum[i] = np.log(epsilon + (re_sum[i] / phi_sum[i]))
 
 
+_nb_optimize_layout_euclidean_single_epoch = numba.njit(
+    _optimize_layout_euclidean_single_epoch, fastmath=True, parallel=False
+)
+
+_nb_optimize_layout_euclidean_single_epoch_parallel = numba.njit(
+    _optimize_layout_euclidean_single_epoch, fastmath=True, parallel=True
+)
+
+
+def _get_optimize_layout_euclidean_single_epoch_fn(parallel: bool = False):
+    if parallel:
+        return _nb_optimize_layout_euclidean_single_epoch_parallel
+    else:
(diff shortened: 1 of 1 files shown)


umap/layouts.py
```diff
@@ -308,9 +324,10 @@ def optimize_layout_euclidean(
     epoch_of_next_negative_sample = epochs_per_negative_sample.copy()
     epoch_of_next_sample = epochs_per_sample.copy()
 
-    optimize_fn = numba.njit(
-        _optimize_layout_euclidean_single_epoch, fastmath=True, parallel=parallel
-    )
+    # Fix for calling UMAP many times for small datasets, otherwise we spend here
+    # a lot of time in compilation step (first call to numba function)
```
kmkolasinski, line 328: I put this comment for myself, but I can remove it as it should not be in the main source code

RzeBar (approved):

lmcinnes:
Good plan, thanks!

kmkolasinski:
wow, that was quick ! when can we expect python package with these changes ?

lmcinnes:
I'm not sure; I would like to try to roll up a bunch of changes if
possible. You can always pip install from the git repo if you need in the
meantime.

j-dutton-alphasights:
Wow, we were just digging into a memory leak we noticed last night,  and came to the conclusion that was is the fix, and just about to propose it - thank you!

kmkolasinski:
hey @lmcinnes  sounds good we will use this approach then. 

 @j-dutton-alphasights  there are probably more functions like this one, which are njit-ed every time we call fit (but they usage depend on the UMAP configuration). I just covered one case with this PR, which is probably also the most common one.

Issue #1082, opened 2023-12-18
Implementation of sciki-learn's get_feature_names_out() API is not correct


buhrmann:
I was happy to see the get_feature_names_out API is already supported, but noticed that the implementation is not correct.

The scikit-learn API for get_feature_names_out(...) accepts optional _input_ feature names, and should generate valid _output_ feature names:

- https://scikit-learn.org/dev/glossary.html#term-get_feature_names_out
- https://scikit-learn.org/dev/modules/generated/sklearn.base.OneToOneFeatureMixin.html#sklearn.base.OneToOneFeatureMixin
- https://scikit-learn.org/dev/modules/generated/sklearn.base.ClassNamePrefixFeaturesOutMixin.html#sklearn.base.ClassNamePrefixFeaturesOutMixin

The current implementation seems to assume that the optional argument passes _output_ names and simply forwards them if provided. This fails in most cases, since scikit-learn pipelines etc. will pass in _input_ names. Specifically, if passed in, this will be an array with as many names as features in the _input_. This is almost never the correct number for the _output_, since UMAP reduces the dimensionality of course. E.g. if I reduce a dataset with 100 original dimensions to 2 final dimensions, the current implementation will forward 100 feature names for an output with only two features.

The simplest fix would be to simply always return a list of names of the correct length, no matter the optional argument.

A slightly more sklearn-ish implementation would be to use the `ClassNamePrefixFeaturesOutMixin`.

PR #1166, opened 2024-11-27
[BUG]: Fix feature names for sklearn API

fraimondo:
This PR fixes the bug described in #1082.

Indeed, current `UMAP` scikit-learn compatible class does not support using pandas dataframes as it will output the input features names.

The fix is to use the `ClassNamePrefixFeaturesOutMixin` and set the `_n_features_out` attribute after fitting.

This will give feature names `umap0`, `umap1`, etc... following what scikit-learn does for the dimensionality reduction transformers.

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -8,8 +8,8 @@ from warnings import warn
 import time
 
 from scipy.optimize import curve_fit
-from sklearn.base import BaseEstimator
-from sklearn.utils import check_random_state, check_array
+from sklearn.base import BaseEstimator, ClassNamePrefixFeaturesOutMixin
+from sklearn.utils import check_array, check_random_state
 from sklearn.utils.validation import check_is_fitted
 from sklearn.metrics import pairwise_distances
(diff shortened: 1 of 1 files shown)


lmcinnes:
This looks good. A couple of catches -- the tests seem to be failing right now; you may have to amend the tests for feature name related things to call "fit" (it currently doesn't), and possibly a few other fixes. The other thing is that dosctrings should be in numpy format if possible. Any chance you can fix this so we can merge this?

fraimondo:
> This looks good. A couple of catches -- the tests seem to be failing right now; you may have to amend the tests for feature name related things to call "fit" (it currently doesn't), and possibly a few other fixes. The other thing is that dosctrings should be in numpy format if possible. Any chance you can fix this so we can merge this?

Just fixed the tests.

Regarding docstrings, I did not make any change that would require modifying or adding docstrings. I'll wait for the CI to see if I'm missing something.

fraimondo:
Ready to merge!

lmcinnes:
Thanks!

PR #1199, opened 2025-04-25
feat: generate embedding from index and initialization

wilsonjr:
This PR aims to reduce memory usage when an index is pre built for the input data. The user is able to generate an embedding given the precomputed index and initialization.

* Generates embeddings using `precomputed_knn` and `init (np.array)`;
* Refactors `fuzzy_simplicial_set` by pulling out `nearest_neighbor` call (so we don't require passing the input data);
* Usage example

### DIFF
--- a/examples/index_fit_example.py
+++ b/examples/index_fit_example.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python
+
+from sklearn.decomposition import PCA
+import umap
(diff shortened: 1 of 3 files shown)


lmcinnes:
This looks quite useful. There are a number of changes, so I'll have to take some time shortly to go through it and verify everything.

PR #1210, opened 2025-07-17
Add reference to torchdr for GPU-accelerated UMAP

huguesva:
Hi Leland and team,

Thank you for the amazing work on this package.

I’d like to add a reference to a GPU‑compatible version of UMAP from the torchdr package. We’ve kept it very faithful to the original UMAP implementation: gradient computation, clipping, and stabilization tricks are all the same. We made it so that every component (kNN, affinity, and embedding optimization) can be accelerated on GPU when available.

I believe this could be a great option for the UMAP community. Please let me know what you think, and feel free to share any feedback if you have a chance to try torchdr UMAP.

Thanks a lot for considering this!

Best,
Hugues

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -462,6 +462,31 @@ Nomic Atlas provides:
 * Rich information display on hover
 * Shareable UMAPs via URL links to your embeddings and data maps in Atlas
 
+
+---------------------------------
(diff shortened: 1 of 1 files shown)


lmcinnes:
Thanks for working on TorchDR. It looks like a great project.

Issue #1235, opened 2026-01-11
Deterministic sort algorithm for fast_knn_indices


bohemian-miser:
Currently the code uses 'quicksort' which does not produce deterministic results. This is causing causing upstream tests to fail in scikit-learn 1.8.0 and likely other places too. Having a deterministic sort would produce reliable results for ties.

PR #1236, opened 2026-01-12
Deterministic sort for fast_knn_indices

bohemian-miser:
This makes tests using UMAP less flaky because ties will be returned in the same order.

This may have a performance impact. If performance is critical here, using np.argpartition would be a better solution (only sorts the first k items) but wouldn't address the deterministic sorting issue.

https://github.com/lmcinnes/umap/issues/1235


bohemian-miser:
How did this result in so many errors?

lmcinnes:
I think I need to fix the azure pipelines yml as this is failing out on that early, so it isn't your changes. It looks like some change at some point messed with the coveralls coverage upload, and it was incorrect, and broke the pipeline. If you resync from master it will hopefully remedy that.

PR #1251, opened 2026-04-11
Fix _select_font_color ValueError on hex backgrounds

### LINKED ISSUE #1231: umap.plot error when using hexadecimal color as background in `_select_font_color()`
Hello,

I've stumbled upon a `ValueError: invalid literal for int() with base 10` in `_select_font_color()`.
In *umap.umap.plot.py* line 241:
```python
[int("0x" + c) for c in (background[1:3], background[3:5], background[5:7])]
```
Simply replace `int("0x" + c)` by `int("0x" + c, 0)` to [interpret the passed string as integer literal](https://docs.python.org/3/library/functions.html#int) to fix the issue:
```python
[int("0x" + c, 0) for c in (background[1:3], background[3:5], background[5:7])]
```
However, I'm not sure if this is some leftover code or if hexadecimal colors are/will be supported.

---
### How this happened
In order to enhance the visibility of a predefined set of label colors I wanted to specify a custom color for the background.
Seeing that `umap.plot.points()` takes a `background` argument, I tried to pass a hexadecimal string since `color_key` works with hexadecimal strings and the docstring states that "any color name will work".
Hence I had hoped that `background` would also work with hexadecimal strings (even though they are admittedly more of a color definition than a color name).
Unfortunately, it resulted in the aforementioned error.

#### Workarounds to specify a custom background color
Simply calling `ax.set_facecolor()` **after** calling `umap.plot.points()` does the trick as long as there are not too many datapoints.  
Otherwise, in the event datashader kicks in, the following monkey patching solution worked for me (since datashader also [supports color tuples](https://datashader.org/api.html#datashader.transfer_functions.set_background)):
```python
class CustomTuple(tuple):
    def __new__(cls, *values):
        return super().__new__(cls, values)

    def __init__(self, *values):
        self.values = values

    def startswith(self, s: str) -> bool:
        """Avoid faulty if-statement."""
        return False

bg = CustomTuple(0, 123, 255) # choose whatever color you need
umap.plot.points(..., ax=ax, background=bg)
```

Kind regards

sebastianbreguel:
Fixes #1231.

`umap.plot._select_font_color` raises `ValueError: invalid literal for int() with base 10: '0xff'` for any hex background color, because `int("0x" + c)` calls `int()` in base 10. Replaced with `int(c, 16)`, which parses the 2-char hex component directly.

```python
from umap.plot import _select_font_color

# before
_select_font_color("#ffffff")
# ValueError: invalid literal for int() with base 10: '0xff'

# after
_select_font_color("#ffffff")
# 'black'
```

### DIFF
--- a/umap/plot.py
+++ b/umap/plot.py
@@ -238,7 +238,7 @@ def _select_font_color(background):
         font_color = "white"
     elif background.startswith("#"):
         mean_val = np.mean(
-            [int("0x" + c) for c in (background[1:3], background[3:5], background[5:7])]
+            [int(c, 16) for c in (background[1:3], background[3:5], background[5:7])]
         )
         if mean_val > 126:
             font_color = "black"


lmcinnes (approved):

lmcinnes:
Nice catch!

PR #1252, opened 2026-04-11
Add py.typed marker for type checkers

### LINKED ISSUE #1177: Request: adding `py.typed` for type checkers
Currently with `umap-learn==0.5.7` and `mypy=1.14.1` I get `import-untyped`:

```none
error: Skipping analyzing "umap": module is installed, but missing library stubs or py.typed marker  [import-untyped]
    from umap import UMAP
    ^
```

It would be nice to add a `py.typed` to `umap` so type checkers like `mypy` can inspect `umap`.

sebastianbreguel:
Fixes #1177.

Adds an empty `umap/py.typed` marker (per PEP 561) and includes it in setuptools `package-data` so it ships with the wheel. Type checkers like `mypy` now recognise `umap` as a typed package instead of emitting:

```
error: Skipping analyzing "umap": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

The package doesn't ship full annotations yet, so this is just the marker — it lets downstream users drop `ignore_missing_imports` overrides for `umap`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,3 +67,4 @@ zip-safe = false
 
 [tool.setuptools.package-data]
 "*" = ["*.pyx", "*.pxd"]
+"umap" = ["py.typed"]


lmcinnes:
I haven't been keeping up with these things. I should really also be going through and adding types to all the function signatures, but that's for another time. I appreciate your work making things at least a little smoother for people using type checkers.

PR #1254, opened 2026-04-11
Remove fastmath=True from layouts.py optimization functions

sebastianbreguel:
## Summary

- Extends the `fastmath` removal from commit 6927806 (`umap_.py`) to `layouts.py`
- `fastmath=True` allows numba to optimize away epsilon guards like `/ (sigmas[k] + 1e-6)`, causing `ZeroDivisionError` in `inverse_transform()`
- Keeps `fastmath=True` only on `rdist()` which is pure squared-distance arithmetic (no divisions)

### Reproducer (from #1228)

```python
import numpy as np
import sklearn.datasets
import umap

data, labels = sklearn.datasets.fetch_openml('mnist_784', version=1, return_X_y=True)
mapper = umap.UMAP(random_state=42).fit(data)

corners = np.array([[-5, -10], [-7, 6], [2, -8], [12, 4]])
test_pts = np.array([
    (corners[0]*(1-x) + corners[1]*x)*(1-y) +
    (corners[2]*(1-x) + corners[3]*x)*y
    for y in np.linspace(0, 1, 10)
    for x in np.linspace(0, 1, 10)
])

inv_transformed_points = mapper.inverse_transform(test_pts)  # ZeroDivisionError
```

Fixes #1228

Issue #1228, opened 2025-11-02
ZeroDivisionError in `inverse_transform()`


jvpeetz:
Dear Leland McInnes,
trying to use the `inverse_transform()` method of umap-learn, following the tutorial on [Inverse transforms](https://umap-learn.readthedocs.io/en/latest/inverse_transform.html), when calling `inv_transformed_points = mapper.inverse_transform(test_pts)`, this exception is hit:

```
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
Cell In[5], line 1
----> 1 inv_transformed_points = mapper.inverse_transform(test_pts)

File /usr/lib/python3/dist-packages/umap/umap_.py:3332, in UMAP.inverse_transform(self, X)
   3329 tail = graph.col
   3330 weight = graph.data
-> 3332 inv_transformed_points = optimize_layout_inverse(
   3333     inv_transformed_points,
   3334     self._raw_data,
   3335     head,
   3336     tail,
   3337     weight,
   3338     self._sigmas,
   3339     self._rhos,
   3340     n_epochs,
   3341     graph.shape[1],
   3342     epochs_per_sample,
   3343     self._a,
   3344     self._b,
   3345     rng_state,
   3346     self.repulsion_strength,
   3347     self._initial_alpha / 4.0,
   3348     self.negative_sample_rate,
   3349     self._inverse_distance_func,
   3350     tuple(self._metric_kwds.values()),
   3351     verbose=self.verbose,
   3352     tqdm_kwds=self.tqdm_kwds,
   3353 )
   3355 return inv_transformed_points

File /usr/lib/python3/dist-packages/umap/layouts.py:850, in optimize_layout_inverse(head_embedding, tail_embedding, head, tail, weight, sigmas, rhos, n_epochs, n_vertices, epochs_per_sample, a, b, rng_state, gamma, initial_alpha, negative_sample_rate, output_metric, output_metric_kwds, verbose, tqdm_kwds, move_other)
    847     tqdm_kwds["disable"] = not verbose
    849 for n in tqdm(range(n_epochs), **tqdm_kwds):
--> 850     optimize_fn(
    851         epochs_per_sample,
    852         epoch_of_next_sample,
    853         head,
    854         tail,
    855         head_embedding,
    856         tail_embedding,
    857         output_metric,
    858         output_metric_kwds,
    859         weight,
    860         sigmas,
    861         dim,
    862         alpha,
    863         move_other,
    864         n,
[...]

PR #1255, opened 2026-04-16
Remove fastmath=True from layouts.py optimization functions

sebastianbreguel:
## Summary

- Remove `fastmath=True` from all `numba.njit()` calls compiling layout optimization epoch functions in `layouts.py` (euclidean, generic, inverse, densmap-init, aligned)
- `fastmath=True` allows LLVM to reorder floating-point operations and ignore NaN/Inf semantics, which can silently produce zero-division or NaN embeddings during `inverse_transform` (ref #1228)
- Add comprehensive benchmark (`benchmarks/bench_fastmath_removal.py`) covering end-to-end fit/transform, isolated epoch micro-benchmarks (fastmath=True vs False), and numerical correctness via trustworthiness scores

## Details

`fastmath=True` is an unsafe Numba/LLVM flag that permits:
- Reassociation of floating-point arithmetic
- Ignoring NaN and Inf propagation rules
- Fused multiply-add transformations that change rounding

In the layout optimization hot loops, distance calculations can produce values very close to zero. With `fastmath=True`, the compiler may reorder operations in a way that skips the `dist > 0` guards, leading to division by zero and NaN embeddings — particularly visible in `inverse_transform`.

The `rdist` helper (line 33) retains `fastmath=True` since it is a pure arithmetic leaf function with no branching or division guards.

Replaces #1254 which was closed when the fork was removed.

## Test plan

- [x] Verify no `fastmath=True` remains in layout optimization `njit()` calls
- [x] Benchmark shows negligible performance impact (<2% in micro-benchmarks)
- [x] Existing test suite passes
- [x] `inverse_transform` no longer produces NaN on edge cases from #1228

cc @lmcinnes

### DIFF
--- a/benchmarks/bench_fastmath_removal.py
+++ b/benchmarks/bench_fastmath_removal.py
@@ -1,0 +1,301 @@
+"""
+Benchmark: fastmath=True removal from layouts.py optimization functions.
+
+Compares performance of UMAP fit/transform with and without fastmath=True
+on the numba-compiled layout optimization functions.
+
+Usage:
+    uv run python benchmarks/bench_fastmath_removal.py
+"""
+
+import time
+import warnings
+
+import numba
+import numpy as np
+from sklearn.datasets import make_blobs, fetch_openml
+
+import umap
+from umap.layouts import (
+    _optimize_layout_euclidean_single_epoch,
+    _optimize_layout_generic_single_epoch,
+    _optimize_layout_inverse_single_epoch,
+)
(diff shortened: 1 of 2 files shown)


sebastianbreguel:
@lmcinnes let me know what you think now!

lmcinnes:
I'll try to run the benchmark script on various local hardware myself to verify today. Thanks for taking the trouble to make the script.

PR #1263, opened 2026-05-26
Vectorize densmap density computation in simplicial_set_embedding()

sebastianbreguel:
## Summary

The density calculation for densMAP (`densmap=True` or `output_dens=True`) iterated over every graph edge in a Python `for`-loop, doing a sparse matrix element access (`dists[j, k]`, O(log d) per lookup) per iteration.

Replaced with a `@numba.njit()` compiled function that does the same loop — no temporary arrays, no extra RAM. Converts `dists` (DOK) to CSR once, then passes the raw CSR arrays (`indptr`, `indices`, `data`) to the numba function for direct compiled iteration.

## Benchmark (10k vertices, 100k edges)

| | Time | Speedup | Peak RAM | RAM delta |
|---|---|---|---|---|
| **Before** | 2,154 ms | — | 0.11 MB | — |
| **After** | 1.6 ms | **1,387× faster (+99.9%)** | 0.11 MB | **0 MB (neutral)** |

## Test plan

- [x] `test_densmap.py` — 3 passed (including `test_densmap_trustworthiness`)
- [x] `test_umap_on_iris.py` — 12 passed
- [x] `test_umap_ops.py` — 19 passed

### DIFF
--- a/umap/umap_.py
+++ b/umap/umap_.py
@@ -935,6 +935,26 @@ def noisy_scale_coords(coords, random_state, max_coord=10.0, noise=0.0001):
     )
 
 
+@numba.njit()
+def _densmap_original_densities(
+    head, tail, graph_data, dists_indptr, dists_indices, dists_data, ro, mu_sum
+):
+    for i in range(len(head)):
+        j = head[i]
+        k = tail[i]
+        d_val = 0.0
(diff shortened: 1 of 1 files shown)


lmcinnes (approved):

lmcinnes:
I swear I have already made that change -- probably in the 0.6dev branch that isn't merged.

PR #1273, opened 2026-05-29
Use argpartition for O(n) selection in fast_knn_indices

sebastianbreguel:
## What

`fast_knn_indices` (`umap/utils.py`) — used by the `fit` precomputed / small-data KNN path (`nearest_neighbors` with `metric="precomputed"`) — fully sorted every distance row with `argsort(kind="mergesort")` (**O(n log n)** per row) and then discarded all but the `n_neighbors` smallest.

This replaces the full sort with `np.argpartition` to select the `k` smallest in **O(n)**, then sorts only those `k` candidates. The candidate indices are pre-sorted ascending before the stable sort, so the result reproduces the previous `argsort(mergesort)` ordering (ties broken by lowest original index) on distinct distances.

This is the `fit`-path counterpart to the `argpartition` selection already used in `UMAP.transform`, so the two paths are now consistent.

## Correctness (output preserved)

- **Distinct distances** (normal continuous data): output is **bit-identical** to the previous implementation — verified on float distance matrices and duplicate-point matrices.
- **Exact-distance ties**: the chosen index among *equidistant* points may differ, but the selected neighbour **distances are identical** — the `k` smallest distances are always returned. Same tradeoff `transform` already ships.
- Stays inside `@numba.njit(parallel=True)` (numba supports `argpartition`).

New regression tests:
- `test_fast_knn_indices_matches_argsort` — pins identical ordering on distinct distances.
- `test_fast_knn_indices_preserves_distances_with_ties` — pins distance preservation under heavy ties.

All existing precomputed / knn / transform tests pass (18 passed, 1 skipped).

## Benchmarks

Symmetric `float32` distance matrix, `n_neighbors=15`. Speed = min of 5×3 runs; RAM = `tracemalloc` peak. Dev machine, single process.

### Speed

| N | Before | After | Δ abs | Δ % | Speedup |
|---|--------|-------|-------|-----|---------|
| 1000 | 5.45 ms | 1.08 ms | −4.37 ms | **−80.2%** | 5.06× |
| 2000 | 21.77 ms | 4.74 ms | −17.02 ms | **−78.2%** | 4.59× |
| 4000 | 101.96 ms | 18.17 ms | −83.79 ms | **−82.2%** | 5.61× |

### Peak RAM

| N | Before | After | Δ abs | Δ % |
|---|--------|-------|-------|-----|
| 1000 | 0.199 MB | 0.184 MB | −0.015 MB | −7.5% |
| 2000 | 0.411 MB | 0.373 MB | −0.038 MB | −9.1% |
| 4000 | 0.819 MB | 0.804 MB | −0.015 MB | −1.9% |

Complexity: **O(n² log n) → O(n²)**. Speed ~4.6–5.6× faster; peak RAM equal-or-lower. Dev-machine numbers; absolute values vary by hardware.

### DIFF
--- a/umap/tests/test_umap_ops.py
+++ b/umap/tests/test_umap_ops.py
@@ -15,7 +15,7 @@ import scipy.sparse
 import pytest
 import warnings
 from umap.distances import pairwise_special_metric
-from umap.utils import disconnected_vertices
+from umap.utils import disconnected_vertices, fast_knn_indices
 from scipy.sparse import csr_matrix
 
 # Transform isn't stable under batching; hard to opt out of this.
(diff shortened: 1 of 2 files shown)


lmcinnes (approved):

lmcinnes:
I seem to recall numba not supporting argpartition once upon a time a long time ago, glad it does now.

PR #1285, opened 2026-08-30
Avoid mutating sparse inputs for empty unions

### LINKED ISSUE #1233: distances are buggy, should be re-exports from pynndescent
The `umap` package seems to reimplement the same distance metrices as the `pynndescent` package. However, the `umap` implementation is buggy in at least one case:

- `sparse.arr_union` can return one of its input arrays
- the return value of `sparse.arr_union` is used as a writable buffer by `sparse.sparse_sum`
- `sparse.sparse_diff` is a thin wrapper around `sparse.sparse_sum`
- `sparse.sparse_diff` is used in `sparse_euclidean`

This leads to two issues:
1. The input array may be modified, which is not expected by the user
2. If `sparse_euclidean`, `sparse_diff`, or `sparse_sum` is called by a custom distance metric, which gets its sparse array via keyword arguments, the keword arguments will be [stored in a closure](https://github.com/lmcinnes/pynndescent/blob/118679c0c3f0fec95f54bd85fbdf63510be7fdab/pynndescent/pynndescent_.py#L849-L856), which leads numba to treat them as readonly, leading to type inference failure (see scverse/muon#173).

The `pynndescent` implementation of `sparse_euclidean` works without any issues. Given that both packages are maintained by you, `umap` depends on and uses `pynndescent`, and the supported distance functions appear to be identical, I think it would make sense if `umap` would simply re-export the distance functions from `pynndescent`.

pttydou:
Fixes #1233.

`arr_union` returned one of its input arrays when the other input was empty. `sparse_sum` uses the union as a writable compaction buffer, which could mutate caller-owned indices and fail for readonly arrays captured by Numba custom metrics.

This copies the nonempty input in the empty-side branches. The both-nonempty path and sparse metric values, dtypes, and ordering remain unchanged.

Tests:
- `python -m pytest umap/tests/test_umap_metrics.py -q` (99 passed, 5 skipped)
- `python -m pytest umap/tests -q -k sparse` (28 passed, 6 skipped, 297 deselected)
- `python -m pytest umap/tests -q --disable-warnings` (265 passed, 66 skipped)
- `python -m black --check --target-version py310 umap/sparse.py umap/tests/test_umap_metrics.py`

### DIFF
--- a/umap/sparse.py
+++ b/umap/sparse.py
@@ -26,9 +26,9 @@ def arr_unique(arr):
 @numba.njit()
 def arr_union(ar1, ar2):
     if ar1.shape[0] == 0:
-        return ar2
+        return ar2.copy()
     elif ar2.shape[0] == 0:
-        return ar1
+        return ar1.copy()
     else:
         return arr_unique(np.concatenate((ar1, ar2)))
 
--- a/umap/tests/test_umap_metrics.py
+++ b/umap/tests/test_umap_metrics.py
@@ -1,5 +1,6 @@
 import numpy as np
-from numpy.testing import assert_array_almost_equal
+import numba
+from numpy.testing import assert_array_almost_equal, assert_array_equal
 import umap.distances as dist
 import umap.sparse as spdist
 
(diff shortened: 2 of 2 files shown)


lmcinnes (approved):

lmcinnes:
Nice catch. Thank you!

